PR #6, opened 2024-09-10
Replace RocksDB by LMDB

Asurar0:
For more information: https://github.com/Sweattypalms/ferrumc/issues/5

Initially, MDBX was considered for implementation over LMDB, but the API of the reference bindings underwent significant changes and support began to decline. As a result, I opted to implement LMDB instead, which benefits from extensive support in Rust libraries.

* Removed RocksDB from Cargo.toml and the database module.
* Added the `heed` crate to Cargo.toml.
* Reimplemented the entire database API to utilize LMDB with tokio blocking tasks.
* Fixed a potential key calculation bug.
* Added documentation.
* Refactored `Database::chunk_exists` to directly load the queried chunk into the cache if it exists.
* Refactored `Database::batch_insert` to avoid querying the database for loading chunks into the cache that the function has already * allocated.

Internal logic changes are denoted by `// WARNING:` comments.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,6 +243,15 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ReCore-sys:
Build fails but that's cos of a broken discord webhook, code looks fine.

PR #22, opened 2024-09-13
Using toolchain config file instead of cmd `rustup default nightly` and fix i32 overflow

NKDark:
Using `rust-toolchain.toml` allows you to set the nightly toolchain for a project rather than globally.
And
Fix the i32 overflow issue.


talves:
This would close #25 

Not sure if there was another issue for this. The error mentioned here is not the error I got, but does solve it with just the toolchain config file.

PR #24, opened 2024-09-13
Add x86_64 macOS & aarch64 Linux auto-build to workflow

43d9ab5:


Sweattypalms:
Why make windows and macOS separate? Can't you just add them to the matrix? with os and target? 
And, what is cross?

43d9ab5:
the file looked nicer like this, and you can easily change things for a "given" system

ReCore-sys:
We really aren't planning on adding proper support for ARM or mac for the time being. We might consider it later but for now the workflow is already pretty cluttered.

PR #27, opened 2024-09-14
fix: making README headings correctly

DeveloLongScript:
some of the headings were not correctly added into the README, this just fixes it


Sweattypalms:
Thanks for that.

PR #30, opened 2024-09-14
Move the server systems to a new task to allow for ctrl+c awaiting

Charterino:


src/main.rs
```diff
@@ -34,9 +26,30 @@ async fn entry() -> Result<()> {
...
+        _ = tokio::signal::ctrl_c() => {
+            info!("Received ctrl+c.. Shutting down..");
+            true
+        }
+    };
 
-    tokio::signal::ctrl_c().await?;
+    if need_to_kill {
+        kill_all_systems().await?;
+    }
```
Asurar0, line 52: Why not just remove `need_to_kill` and put `kill_all_systems().await?` under `_ = tokio::signal::ctrl_c()` ?

Asurar0 (approved):

src/main.rs
Charterino: You cant have blocks of different types in select! and no two async blocks ever have the same type.

PR #34, opened 2024-09-15
Added default encoding for login_disconnect, login_play, outgoing ping, and outgoing status

en1gma713:


Sweattypalms:
Thanks so much :)

PR #42, opened 2024-09-16
feat: add accurate player list and count to server status

lwalton101:


Sweattypalms:
You could use the ECS and directly run a query on the players! Please update it to do it like, would be must appreciated! If you need some inspiration, the chunk sender could of an example of how to query the ECS.

lwalton101:
Thanks for that, this should be much cleaner now 👍

Sweattypalms:
Alright thanks so much!

PR #52, opened 2024-09-17
Add rsa keypairs with correct (ans.1 der) encoding

en1gma713:
basically add a sane way to generate RSA keys and encrypt/decrypt based on those keys

### DIFF
--- a/src/net/encryption/src/key.rs
+++ b/src/net/encryption/src/key.rs
@@ -1,0 +1,42 @@
+use rsa::{RsaPublicKey, RsaPrivateKey, pkcs1::EncodeRsaPublicKey, Pkcs1v15Encrypt};
+use rand::rngs::OsRng;
+
+pub struct KeyPair {
+    pub public_key: RsaPublicKey,
+    pub encoded_public_key: der::Document,
+    pub private_key: RsaPrivateKey,
+}
+
+impl KeyPair {
+    pub fn new() -> Self {
+        let mut rng = OsRng;
+        let private_key = RsaPrivateKey::new(&mut rng, 1024).expect("Failed to generate a key");
+        let public_key = RsaPublicKey::from(&private_key);
+        let encoded_public_key = public_key.to_pkcs1_der().expect("Failed to encode public key");
+
+        Self {
+            public_key,
+            encoded_public_key,
+            private_key,
+        }
+    }
+
+    pub fn encrypt (&self, data: &Vec<u8>) -> Vec<u8> {
+        let mut rng = OsRng;
+        self.public_key.encrypt(&mut rng, Pkcs1v15Encrypt, &data).expect("Failed to encrypt data")
+    }
+
+    pub fn decrypt (&self, data: &Vec<u8>) -> Vec<u8> {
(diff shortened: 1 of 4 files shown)


src/net/encryption/src/key.rs
```diff
@@ -0,0 +1,42 @@
...
+            private_key,
+        }
+    }
+
+    pub fn encrypt (&self, data: &Vec<u8>) -> Vec<u8> {
+        let mut rng = OsRng;
+        self.public_key.encrypt(&mut rng, Pkcs1v15Encrypt, &data).expect("Failed to encrypt data")
+    }
+
+    pub fn decrypt (&self, data: &Vec<u8>) -> Vec<u8> {
```
Asurar0, line 29: Why using `&Vec<u8>` instead of `&[u8]` ? This could potentially force us to do reallocation.

src/net/encryption/src/key.rs
```diff
@@ -0,0 +1,42 @@
+use rsa::{RsaPublicKey, RsaPrivateKey, pkcs1::EncodeRsaPublicKey, Pkcs1v15Encrypt};
+use rand::rngs::OsRng;
+
+pub struct KeyPair {
+    pub public_key: RsaPublicKey,
+    pub encoded_public_key: der::Document,
+    pub private_key: RsaPrivateKey,
+}
```
Asurar0, line 8: Some small documentation would be appreciated here

Asurar0 (changes requested):

src/net/encryption/src/key.rs
en1gma713: I’ll fix the type and add some comments later tonight

src/net/encryption/src/key.rs
```diff
@@ -0,0 +1,42 @@
...
+        Self {
+            public_key,
+            encoded_public_key,
+            private_key,
+        }
+    }
+
+    pub fn encrypt (&self, data: &Vec<u8>) -> Vec<u8> {
+        let mut rng = OsRng;
+        self.public_key.encrypt(&mut rng, Pkcs1v15Encrypt, &data).expect("Failed to encrypt data")
```
nulluser0, line 26: Maybe for error handling, add error variants to `errors.rs` and instead of `.expect()`, use `.map_err()?` or `.ok_or()?` ?

Other than that, everything looks good and i agree with Asurar0's comments

nulluser0 (changes requested):

en1gma713:
Should be fixed and errors handled now

nulluser0 (approved):
Looks good imo

ReCore-sys:
ill merge into the current code for now and we can move it over when the rewrite is complete

ReCore-sys:
nevermind i didn't read the code right lol, i though it was for the old code

ReCore-sys:
conflicts in
src/net/encryption/Cargo.toml
src/net/encryption/src/lib.rs
@en1gma713
fix these and I'll merge it

en1gma713:
the rewrite's entire file structure updated before i could get this in. closing this and submitting a new one with the correct structure.

PR #61, opened 2024-10-11
Feat: Dockerfile addition + Example World

cappuccinocosmico:
Example world generated in 1.20.1 with seed "ferrumc". Only takes up 100mb in the current version, if its to much I can always move it to a cdn and scrub the git history.


Sweattypalms:
We're gonna make the rewrite on 1.21.1 or whatever the latest version is, could u update it to that version? And 100mb.... hmm... maybe we could ask and download at runtime?

cappuccinocosmico:
Force pushed the removal of all the files, currently its using some infra from my main job, but we can refactor that once we get a blog up and running.

PR #64, opened 2024-10-15
Feature gating the database backends

ReCore-sys:
Due to rocksdb's insanely long compile time, I have added a rocksdb feature and required it to be enabled for rocksdb to be compiled and used. Since it's off by default it will essentially disable rocksdb integration unless explicitly enabled.

### DIFF
--- a/src/lib/storage/src/benches/databases.rs
+++ b/src/lib/storage/src/benches/databases.rs
@@ -71,7 +71,8 @@ pub fn database_benchmarks(c: &mut Criterion) {
             temps.push(db_file);
             backend
         });
-
+        
+        #[cfg(feature = "rocksdb")]
         let mut rocksdb_backend = runtime.block_on(async {
             let mut db_file = PathBuf::from("R:/temp/rocksdb");
             if !db_file.exists() {
@@ -124,6 +125,7 @@ pub fn database_benchmarks(c: &mut Criterion) {
                 });
             },
         );
+        #[cfg(feature = "rocksdb")]
         write_group.bench_with_input(
             "RocksDB",
             &("test".to_string(), data.clone()),
@@ -149,6 +151,7 @@ pub fn database_benchmarks(c: &mut Criterion) {
             runtime
                 .block_on(sled_backend.insert("test".to_string(), key, data.clone()))
                 .unwrap();
+            #[cfg(feature = "rocksdb")]
             runtime
                 .block_on(rocksdb_backend.insert("test".to_string(), key, data.clone()))
                 .unwrap();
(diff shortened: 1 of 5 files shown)


ReCore-sys:
Nah having the rest of them is fine, rocksdb just has extremely long compile times and gets in the way of development and CI

GStudiosX2:
> Nah having the rest of them is fine, rocksdb just has extremely long compile times and gets in the way of development and CI

Well this would allow for people to compile stuff they actually need though and have a smaller exe

ReCore-sys:
But it would mean you are locked into a single database choice unless you have the technical skills to enable the features and compile it yourself

GStudiosX2:
> But it would mean you are locked into a single database choice unless you have the technical skills to enable the features and compile it yourself

No? You can enable both surrealkv and sled by default I mean your currently disabling one of the options by default anyway

Skullians:
Just set up a Jenkins server for CI
then no GitHub limits

GStudiosX2:
https://github.com/marketplace/actions/rust-cache maybe this might help though since it probably would cache rocksdb and not require to recompile every time

Skullians:
Just use a CI server 😭😭😭

Skullians:
Yeah until your limits run out because CERTAIN people don’t skip ci stupid shit like readme changes

ReCore-sys:
And instead we have to pay money to run the server?

ReCore-sys:
Yeah, a very low powered one that really doesn't have the power to run compilations. Also currently we don't need it. If small things like the readme are being updated, the CI won't have to recompile anything so all it'll do is run tests which is pretty fast. And the PR process stops people from making lots of small changes over several commits, so it's really not an issue right now. If we start running into issues, we can explore our options but for now it's really not needed.

radstevee (approved):
I think this would be great to have due to me sometimes working on Codespaces with low-power CPUs, thus taking hours to compile rocksdb.

src/lib/storage/src/benches/databases.rs
```diff
@@ -71,7 +71,8 @@ pub fn database_benchmarks(c: &mut Criterion) {
             temps.push(db_file);
             backend
         });
-
+
```
radstevee, line 74: All of this could just be in one, feature-gated block.
```rs
#[cfg(feature = "rocksdb")]
{
    let mut rocksdb_backend = ...;
    // Everything else
}

Skullians (commented):
looks amazing, definitely will need this for when I never contribute kek

Sweattypalms (approved):
Looks good. Will definitely help with the compilation times, considering rocksdb takes a lot of time.

PR #67, opened 2024-10-16
Start of login Sequence (very basic)

lwalton101:
Packets implemented so far:
- [x] Login Start (was already mostly finished)
- [x] Login Success S2C
- [x] Login Acknowledged C2S

I was going to attempt the encryption packets however there seems to be another pull request in progress so I have only been doing the required packets.


Sweattypalms:
Thanks mate.

PR #75, opened 2024-10-20
remove config creation prompt

radstevee:
This option is practically unneeded since why would you not need a config?


ReCore-sys:
Has been added by https://github.com/ferrumc-rs/ferrumc/pull/77

PR #84, opened 2024-10-23
Config simplification

ReCore-sys:
Simplifying the config system

## Description
The config system was overly complicated and contained a lot of clutter that made it difficult to decipher. This PR changes the config init system from a complex OnceLock system to a much more streamlined lazy_static system. The change from runtime init to a "life before main" system is minor since the config would be one of the first things set up anyway. The only major issue would be not being able to use tracing since that won't have been initialized yet.
The other change implemented was changing how the favicon is fetched, switching it from a runtime loading to simply using lazy_static to bake the favicon directly into the binary.

## Motivation and Context
This PR makes the config much easier to work with and debug. It was discussed with members in the discord server.

## How has this been tested?
It has been tested manually since unit tests of this nature would be largely incompatible with CI

## Types of changes

## Checklist:


.etc/example_config.toml
```diff
@@ -0,0 +1,12 @@
+host = "0.0.0.0"  # IP or hostname where the server will bind
+port = 25565  # Server port (0-65535)
+motd = ["Welcome to the server!", "Have a great time!", "Adventure awaits!"]  # Message of the day
+max_players = 100  # Maximum number of players
+network_tick_rate = 20  # Network updates per second per user
+world = "World"  # The name of the world to load
```
GStudiosX2, line 6: Wasn't the previous value for this `world = "world"`?

ReCore-sys: yes, but it doesn't really matter

GStudiosX2: > yes, but it doesn't really matter

Yeah I guess it's not used yet still think it should be the same as it was before thougt

PR #86, opened 2024-10-25
feature/better_ecs

Sweattypalms:
like. i just. it works. that's what matters. dont question the `*const ()`. Oh and the server doesn't fucking die with more than 10 players 💀💀💀💀


Sweattypalms:
oopsie doopsie

PR #105, opened 2024-10-29
NBT -> Struct mappings + Benches

Sweattypalms:


Sweattypalms:
Also moved main parser file to its own sub module

PR #106, opened 2024-10-29
Accurate player list

Haolaaa:


radstevee (approved):

Sweattypalms:
Please use `into_entities` on the query. And to get the count, use the Vec's length. And for displaying, use the first 5 entries or only show top few players. Since with more players it'd definitely become a mess. You can loop over the `Vec<usize>` where `usize` is the Entity Id. Just do `state.universe.get::<PlayerIdentity>(entity)` and collect it.

Haolaaa:
Done! Thanks for the feedback, let me know if there’s anything else to adjust!

Sweattypalms:
Please merge the main branch with your branch so these clippy warnings go away! Sorry!

PR #115, opened 2024-11-03
Moved the state to it's own crate

ReCore-sys:
This PR moves the ServerState and GlobalState types into it's own crate

## Description
The definitions for the GlobalState and ServerState have been moved into it's own crate since it made very little sense to have it in the net crate. On the surface this seems like it would cause cyclic dependency issues but it won't actually introduce any more issues than having it in the net crate, other than preventing the net crate from providing types that are used in the state, but it wasn't doing that in the first place.

## Motivation and Context
This change was implemented to clear up confusion regarding having the state implemented in the net crate.

## How has this been tested?
Since it's just moving definitions from one place to another, no unit tests are required because if it compiles, it works.

## Types of changes

## Checklist:


Skullians (approved):
slayed

PR #117, opened 2024-11-05
added a 100ms sleep to the keep alive system to prevent it from having excessive cpu usage

AnonymousBit0111:


Sweattypalms:
^ There is no point checking keepalive every 100ms. Since keepalive typically runs every 30 seconds, checking every 15 seconds ensures we catch all players' status even if they join right after a check, while still being far more efficient than checking every 100ms.

src/bin/src/systems/keep_alive_system.rs
```diff
@@ -84,10 +85,11 @@ impl System for KeepAliveSystem {
             if let Err(e) = state.broadcast(&packet, broadcast_opts).await {
                 error!("Error sending keep alive packet: {}", e);
             };
-
+            // TODO, this should be configurable as some people may have bad network so the clients may end up disconnecting from the server moments before the keep alive is sent
```
Sweattypalms, line 88: 15 seconds delay means the client gets 2 chances. The default timeout for keepalive is 30 seconds.

AnonymousBit0111: Ok I have no issue changing this since we will be making it configurable later, in the meantime though, there's an issue where the keep alive runs twice every 30 seconds or 15, whatever the timeout is configured to be, can you look into it, I don't really get why its doing this, everything should be done synchronously in the loop so I'm very lost.

Skullians (changes requested):
Did you mean to type 30?

Skullians (approved):
LGTM
- professional rust dev

AnonymousBit0111:
#119 should be fixed now can you please check @0xnim , I'm not sure how to simulate a client timing out

0xnim:
@AnonymousBit0111 another issue has occured, only a single player can join

src/lib/net/src/packets/incoming/keep_alive.rs
```diff
@@ -13,15 +15,42 @@ pub struct IncomingKeepAlivePacket {
...
+                "Invalid keep alive packet received from entity {:?} with id {:?} (expected {:?})",
                 conn_id, self.id, last_keep_alive.id
             );
+            return NetResult::Err(crate::errors::NetError::Packet(
+                crate::errors::PacketError::InvalidState(0x18),
+            ));
             // TODO Kick player
+        }
+
+        let result = state.universe.get_mut::<IncomingKeepAlivePacket>(conn_id);
```
0xnim, line 32: This is where the problem happens I'm guessing.

src/lib/net/src/packets/incoming/keep_alive.rs
```diff
@@ -13,15 +15,42 @@ pub struct IncomingKeepAlivePacket {
...
+                    err
+                );
+                return Err(crate::errors::NetError::ECSError(result.err().unwrap()));
+            }
         } else {
-            *last_keep_alive = KeepAlive::from(self.id);
+            let mut last_received_keep_alive: ComponentRefMut<'_, IncomingKeepAlivePacket> =
+                result.unwrap();
+
+            *last_received_keep_alive = self;
```
Sweattypalms, line 53: Why not just do a match statement?
```rs
match result {
    Ok(component) => {
        // ...
    },
    Err(e) => {
        // ...
    }
}
```

0xnim: doesnt fix the problem of the error?

0xnim:
```rs
state.universe.get_mut::<IncomingKeepAlivePacket>(conn_id)
```
 is the culprit

0xnim: Can this be a deadlock from dashmap?

Sweattypalms: If something stupid is being done, probably. I haven't read the code in too much depth, but this shouldn't be the case.

AnonymousBit0111: it wasnt, I think the error system is a bit inconsistent, I expected only ComponentTypeNotFound, but also got ComponentRetrievalError, when they're both reporting what should be the same error since the IncomingKeepAlive hasn't been created yet, I've fixed this in #122

PR #122, opened 2024-11-11
fixed only one client being able to join the server

AnonymousBit0111:
- I have tested this with three clients aswell, there seems to be no issue


AnonymousBit0111:
closed as ive restructured the pr in #126

PR #125, opened 2024-11-14
Nix flake

nicholaschiasson:
* Adding initial nix flake capable of building the code
* Checking Cargo.lock into version control
  * This is kind of necessary for the flake to build, but if there is a very good reason you did not have the file checked in, then we can try to find a way around that for the flake
* Room to improve nix flake dev shell


0xnim:
Do you know what will happen if cargo lock is removed?

0xnim:
I'm guessing nothing because it builds to a binary at the end. But just to be sure.

0xnim:
There is no use for nix in this repository.

Instead I have come up with a solution for you to be able to contribute while using a flake.nix.

You can run

`git add --intent-to-add flake.nix`
`git update-index --assume-unchanged flake.nix`

So you don't have to unstage flake.nix everytime you commit.

On the case that you might need to use `git pull --rebase`. You should stash flake.nix and just unstash it after the rebase.

Eveeifyeve:
> There is no use for nix in this repository.
> 
> Instead I have come up with a solution for you to be able to contribute while using a flake.nix.
> 
> You can run
> 
> `git add --intent-to-add flake.nix` `git update-index --assume-unchanged flake.nix`
> 
> So you don't have to unstage flake.nix everytime you commit.
> 
> On the case that you might need to use `git pull --rebase`. You should stash flake.nix and just unstash it after the rebase.

As a nix/nixos user it would be great to have a devshell that is provided by deafult and I will be happily able to provide a cleaner approach to this which supports backwards compatibility with shell.nix using flake-compact and supports flake usage.

PR #126, opened 2024-11-16
Feat/kicking inactive

AnonymousBit0111:
This PR kicks inactive players and remove them from the universe

## Description
Keep alives are now stored in the Incoming and Outgoing Keep alive packet, if the last incoming keep alive is not the same as the last outgoing one the client should be kicked and if a client fails to send a keep alive for 30 its also kicked, also now if it takes more than 2 seconds to read a packet the connection loop will skip to the next iteration to prevent the loop from hanging.

## Motivation and Context
This change was implemented to make sure inactive or invalid clients are kicked

## How has this been tested?
Ive connected to the server with two clients and disconnected them by turning off their internet, if I turn the interned back on within 30 seconds nothing happens, otherwise the client is kicked.

Screenshots (if appropriate):
Types of changes

- [x]  Bug fix (non-breaking change which fixes an issue)
Checklist:
- [x]  My code follows the code style of this project.
- [ ]  I have added tests to cover my changes.


AnonymousBit0111:
someone help me with these merge conflicts please, it should take everything from my branch and discard everything otherwise and I'm not quite sure how to do that

AnonymousBit0111:
dont merge yet, seems like because of the conflicts now it kicks even when the keep alive has been received

src/bin/src/systems/keep_alive_system.rs
AnonymousBit0111, line 1: remove traces when random stop issue is fixed

AnonymousBit0111:
looks like a deadlock...

PR #128, opened 2024-11-17
Fix/keep alive stuff

AnonymousBit0111:
This PR kicks inactive players and remove them from the universe

Description
Keep alives are now stored in the Incoming and Outgoing Keep alive packet, if the last incoming keep alive is not the same as the last outgoing one the client should be kicked and if a client fails to send a keep alive for 30 its also kicked, also now if it takes more than 2 seconds to read a packet the connection loop will skip to the next iteration to prevent the loop from hanging.

Motivation and Context
This change was implemented to make sure inactive or invalid clients are kicked

How has this been tested?
Ive connected to the server with two clients and disconnected them by turning off their internet, if I turn the interned back on within 30 seconds nothing happens, otherwise the client is kicked.

Screenshots (if appropriate):
Types of changes

- [x]  Bug fix (non-breaking change which fixes an issue)

Checklist:
 

- [ ]  I have added tests to cover my changes.
 - [x]  Clippy passes with no warnings.


src/bin/src/systems/keep_alive_system.rs
```diff
@@ -46,19 +48,17 @@ impl System for KeepAliveSystem {
...
                 .into_entities()
                 .into_iter()
                 .filter_map(|entity| {
                     let conn_state = state.universe.get::<ConnectionState>(entity).ok()?;
-                    let keep_alive = state.universe.get_mut::<KeepAlive>(entity).ok()?;
+                    let keep_alive = state.universe.get_mut::<IncomingKeepAlive>(entity).ok()?;
 
                     if matches!(*conn_state, ConnectionState::Play)
-                        && (current_time - keep_alive.id) >= fifteen_seconds_ms
+                        && (current_time - keep_alive.id) >= 15000
```
saadm-itc, line 61: Could you rename the field `id` to `timestamp` if that's what it is? Would be easier to understand this code easily in the future. Since this seems quite a weird calculation at the moment by looking at the code.

Sweattypalms, line 61: Could you rename the field `id` to `timestamp` instead? Since I feel like it'd make more sense in the future when looking at this code. Since right now it looks like a quite weird calculation between an id and timestamp :)

src/bin/src/systems/keep_alive_system.rs
```diff
@@ -68,26 +68,58 @@ impl System for KeepAliveSystem {
...
-            let packet = KeepAlivePacket::default();
+                // I know this is the second iteration of the entities vector, but it has to be done since terminate_connection is async
+                for entity in entities.iter() {
+                    let keep_alive = state
+                        .universe
+                        .get_mut::<IncomingKeepAlive>(*entity)
+                        .ok()
+                        .unwrap();
+
+                    if (current_time - keep_alive.id) >= 30000 {
```
Sweattypalms, line 80: Here are well. `id` -> `timestamp`

src/lib/net/src/packets/incoming/keep_alive.rs
```diff
@@ -1,27 +1,33 @@
...
+use crate::utils::state::terminate_connection;
 use crate::{NetResult, ServerState};
 use ferrumc_macros::{packet, NetDecode};
 use std::sync::Arc;
 use tracing::debug;
 
 #[derive(NetDecode)]
 #[packet(packet_id = 0x18, state = "play")]
-pub struct IncomingKeepAlivePacket {
+pub struct IncomingKeepAlive {
```
Sweattypalms, line 11: Please keep packets have a prefix `Packet` since they're publicly exported and it'd be confusing.

src/lib/net/src/packets/outgoing/disconnect.rs
```diff
@@ -0,0 +1,27 @@
+use ferrumc_macros::{packet, NetEncode};
+use ferrumc_text::{ComponentBuilder, TextComponent};
+use std::io::Write;
+
+#[derive(NetEncode)]
+#[packet(packet_id = 0x1D)]
+pub struct Disconnect {
```
Sweattypalms, line 7: Same here. Suffix with `Packet`

src/lib/net/src/utils/state.rs
```diff
@@ -0,0 +1,65 @@
...
+            conn_id, e
+        );
+        return Err(e);
+    }
+
+    match conn_id.get_mut::<ConnectionControl>(state.clone()) {
+        Ok(control) => {
+            control
+                .should_disconnect
+                .store(true, std::sync::atomic::Ordering::Relaxed);
```
Sweattypalms, line 52: Why is there an atomic inside of a component? Components' are already protected with read-write locks.

AnonymousBit0111: didnt know, will change

Sweattypalms:
thanks bbg

PR #129, opened 2024-11-27
WIP: Chunk Sending

ReCore-sys:
Adds functionality for sending chunks

## Description
This PR adds simple chunk sending and lays out the groundwork for fully fledged world loading. It adds the required packets and such for sending the world to a client. There is a *lot* of room for improvement, but it should be enough for now to explore imported worlds. The current system just sends a 10x10 section of the world to all players every 5 seconds, but this will need to heavily modified since it is currently extremely inefficient and falls apart very quickly. It does also have some issues with not showing some chunks but i think this is just due to a bad method of sending chunks and should be fixed when overhauled.
Overall this is currently a very brittle system, but should suffice until I have a chance to improve it.

## Motivation and Context
This links to https://github.com/ferrumc-rs/ferrumc/issues/70 and is a vital component of the overall server. This also marks the end of the V3 rewrite stage and a transition should occur shortly after.

## How has this been tested?
Since it is largely networking based, unit tests were not practical, but it was tested extensively manually. It was primarily tested with the [Fabulously Optimized](https://modrinth.com/modpack/fabulously-optimized/version/6.2.3) (Version 6.2.3) modpack which does use an optimized Sodium renderer mod, so it may not achieve 1:1 parity with the vanilla client in regards to the logic used client side for rendering, but since Sodium doesn't change any of the networking, there shouldn't be any noticeable differences.

image_2024-11-27_143649834

## Types of changes

## Checklist:


0xnim:
Chunks do not send outside +-12x12. Interesting

src/lib/net/Cargo.toml
```diff
@@ -28,3 +29,4 @@ serde_derive = { workspace = true }
 rand = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
 async-trait = { workspace = true }
+bitreader = { workspace = true }
```
0xnim, line 32: Is this ever used?

ReCore-sys: Nah, I thought I needed to be able to read the packed integers, but I ended up not needing it and I guess I forgot to remove the import.

0xnim:
> Chunks do not send outside +-12x12. Interesting

https://wiki.vg/Protocol#Set_Center_Chunk 

Needs to be implemented before merge.

src/lib/net/src/packets/outgoing/chunk_and_light_data.rs
```diff
@@ -0,0 +1,169 @@
...
+            empty_sky_light_mask,
+            empty_block_light_mask,
+            sky_light_arrays: LengthPrefixedVec::new(sky_light_arrays),
+            block_light_arrays: LengthPrefixedVec::new(block_light_arrays),
+        }
+    }
+    
+    
+    pub async fn from_chunk(chunk: &Chunk) -> Result<Self, NetError> {
+        let mut data = Cursor::new(Vec::new());
```
0xnim, line 66: Unsure, but I dont think Cursor is Async.

0xnim (changes requested):
Just the changes I mentioned. And the set chunk center. Otherwise great.

src/lib/net/src/packets/outgoing/chunk_and_light_data.rs
ReCore-sys: No, but the AsyncWriter trait works on it and allows me to write more specific data types like u16s

src/bin/src/systems/chunk_sender.rs
```diff
@@ -0,0 +1,78 @@
+use crate::systems::definition::System;
```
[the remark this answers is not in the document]
ReCore-sys: Yeah 100%, but the entire thing needs to be redone with optimizations anyway, so I'll do that when I redo the rest of it anyway.

src/lib/net/src/packets/outgoing/set_center_chunk.rs
0xnim, line 1: I wrote part of this so someone else check it @Sweattypalms

src/lib/net/src/packets/outgoing/set_render_distance.rs
0xnim, line 1: I wrote this so Sweaty check it @Sweattypalms

Sweattypalms: hi

src/bin/src/systems/chunk_sender.rs
```diff
@@ -0,0 +1,132 @@
...
+impl System for ChunkSenderSystem {
+    async fn start(self: Arc<Self>, state: GlobalState) {
+        info!("Chunk sender system started");
+
+        while !self.stop.load(Ordering::Relaxed) {
+            let players = state
+                .universe
+                .query::<(&PlayerIdentity, &Position, &mut StreamWriter)>();
+            // TODO: This is so ass. Please fix this.
+            for (_entity, (player, position, mut conn)) in players {
```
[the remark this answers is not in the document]
Sweattypalms: And like I mentioned, don't get a `mut conn` lock here. It will block every other operation that requires the connection. Instead use the `entity` (id) and get the `<StreamWriter>` when you're ready to send the chunks.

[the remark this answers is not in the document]
Sweattypalms: Another thing is, this will eventually be not very useful. We will need to implement a system that dynamically sends chunks based on `Δ pos`. Like we did in v2. If the last chunk send position vs now is a threshold apart, then send the chunks.

src/bin/src/systems/tcp_listener_system.rs
```diff
@@ -29,12 +29,11 @@ impl TcpListenerSystem {
     async fn initiate_loop(state: GlobalState) -> Result<()> {
         let tcp_listener = &state.tcp_listener;
         info!("Server is listening on [{}]", tcp_listener.local_addr()?);
-
-
+        
         loop {
+            info!("Accepting connection");
```
Sweattypalms, line 34: `info!` rahhh. Just use debug here, and when connection is established (when we have remote addr) then send `info` like "Accepted connection from {:?}" or something similar.

Sweattypalms (changes requested):

src/lib/world/src/db_functions.rs
```diff
@@ -52,6 +56,23 @@ pub(crate) async fn load_chunk_internal(
...
+    world.storage_backend.batch_get("chunks".to_string(), digests).await?.iter().map(|chunk| {
+        match chunk {
+            Some(compressed) => {
+                let data = world.compressor.decompress(compressed)?;
+                let chunk: Chunk = bitcode::decode(&data).map_err(|e| WorldError::BitcodeDecodeError(e.to_string()))?;
+                Ok(chunk)
+            }
+            None => Err(WorldError::ChunkNotFound),
+        }
+    }).collect()
```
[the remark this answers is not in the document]
ReCore-sys: Batches are faster yeah

src/lib/net/src/packets/outgoing/chunk_and_light_data.rs
```diff
@@ -0,0 +1,183 @@
...
+            data: LengthPrefixedVec::new(vec![0; SECTIONS * 10]),
+            block_entities: LengthPrefixedVec::new(Vec::new()),
+            sky_light_mask: BitSet::new(SECTIONS),
+            block_light_mask: BitSet::new(SECTIONS),
+            empty_sky_light_mask,
+            empty_block_light_mask,
+            sky_light_arrays: LengthPrefixedVec::new(sky_light_arrays),
+            block_light_arrays: LengthPrefixedVec::new(block_light_arrays),
+        }
+    }
```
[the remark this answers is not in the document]
ReCore-sys: As I said, AsyncWriter lets me write specific data types, whereas the sync versions don't. And right now const optimizations would be a little faster, but once the lighting is generated dynamically we won't be able to use const stuff as much, so i left it this way to avoid restructuring later down the line.

src/bin/src/systems/chunk_sender.rs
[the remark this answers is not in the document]
ReCore-sys: Yeah I'm aware and have said several times. This is a shit system and will need to be redone, but it does an ok enough job for now, but I just wanna finish up the rewrite and fixing this will push it back a while.

src/bin/src/systems/tcp_listener_system.rs
ReCore-sys: sure

src/lib/net/src/packets/outgoing/chunk_and_light_data.rs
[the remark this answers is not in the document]
Sweattypalms: Byteorder is not needed. NetEncode trait does bigendian already!!!

src/lib/net/src/packets/outgoing/chunk_and_light_data.rs
```diff
@@ -89,51 +89,47 @@ impl ChunkAndLightData {
             };
             block_light_data.push(section_block_light_data);
 
-            data.write_u16(section.block_states.non_air_blocks).await?;
+            data.write_u16::<BigEndian>(section.block_states.non_air_blocks)?;
```
Sweattypalms, line 92: U can do .encode directly on non_air_blocks

src/lib/net/src/packets/outgoing/chunk_and_light_data.rs
```diff
@@ -89,51 +89,47 @@ impl ChunkAndLightData {
...
-                    palette_entry.write_async(&mut data).await?;
+                    palette_entry.write(&mut data)?;
                 }
 
                 // Write the data
-                VarInt::new(section.block_states.data.len() as i32)
-                    .write_async(&mut data)
-                    .await?;
+                VarInt::new(section.block_states.data.len() as i32).write(&mut data)?;
                 for data_entry in &section.block_states.data {
```
Sweattypalms, line 107: Can use simd here too. From ferrumc general purpose lib

src/lib/net/src/packets/outgoing/chunk_and_light_data.rs
```diff
@@ -89,51 +89,47 @@ impl ChunkAndLightData {
...
-                    data.write_i64(*data_entry).await?;
+                    data.write_i64::<BigEndian>(*data_entry)?;
```
Sweattypalms, line 108: Don't use byteorder, all primitives implement `NetEncode` trait. Use it.

Sweattypalms (changes requested):

src/lib/net/src/packets/outgoing/chunk_and_light_data.rs
ReCore-sys: they compile to the same thing and it is easier to understand what is going on, since it's not writing to a socket stream, using netencode seems misleading. There isn't a performance difference and it makes it clearer that it's not writing straight to a socket.

PR #138, opened 2024-12-10
Less terrible chunk sending

ReCore-sys:
This PR fixes the atrocious chunk sending system I implemented before. Turns out simply hurling over 500 chunks at a client every 5 seconds was a sub-optimal strategy, who would have guessed?

## Description
This PR implements a smarter "on-demand" chunk sending system to avoid overloading both the server and the client. The previous system had many flaws, including huge lag spikes for both the client and the server, extremely delayed chunk loading times, players only receiving world data every 5 seconds leading to extremely long join times, loading and sending heaps of unneeded data and unnecessarily prolonged component locking which has the potential to slow down the entire server. This new system splits the work over many tasks/threads and doesn't fetch or send unnecessary data, massively reducing the load on both the client and the server.
It works by attaching a `ChunkReciever` component to players and then using that to calculate what chunks the client needs whenever they cross a chunk border, meaning that clients will almost immediately receive the chunks they just came into view range of and also makes sure they aren't sent chunks they can already see. It also splits the work as much as it can by having 2 ECS systems used, one that fetches the chunks from the database and one that sends those chunks to the client. It also uses a separate tokio task for each player when fetching or sending chunk, splitting the work up even further. This approach is explained in more detail in the docs (WIP).
It is important to note that this PR doesn't add support for chunks being updated or chunks being sent over multiple ticks, those will need to be addressed in future PRs.

## Motivation and Context
The previous chunk sending system worked (somewhat) and was merged to just get something quick and dirty out the door as soon as possible. This PR massively improves pretty much everything about it, and should hold up for the foreseeable future.

## How has this been tested?
These changes are hard to test with unit tests, but has been manually tested extensively. Some deadlocking issues did crop up and while I'm pretty confident they have been fixed, I was unable to reproduce them reliably so there is a chance they might persist.

## Types of changes

## Checklist:


src/lib/core/src/chunks/chunk_receiver.rs
```diff
@@ -0,0 +1,46 @@
+use dashmap::{DashMap, DashSet};
+use ferrumc_world::chunk_format::Chunk;
+use tokio::time::Instant;
+
+const VIEW_DISTANCE: i32 = 8;
```
WhaleFromMars, line 5: the view distance is hard-coded, add it to and pull from config, if possible have some reference to the player so in future we can grab their client-side view distance: https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Protocol#Client_Information_(configuration) so we arent sending them chunks they immediately discard if their distance is smaller than config distance

PR #140, opened 2024-12-11
Feature: Whitelist

WhaleFromMars:
adds whitelist support - ~~same format as the vanilla file~~

I can look at adding tests afterwards for valid or not configs, all been manually tested so far


src/lib/core/src/identity/player_identity.rs
```diff
@@ -9,3 +17,30 @@ impl PlayerIdentity {
...
+where
+    D: serde::Deserializer<'de>,
+{
+    let uuid_str: String = Deserialize::deserialize(deserializer)?;
+    let id = uuid::Uuid::try_parse(&uuid_str)
+        .map_err(|e| serde::de::Error::custom(format!("failed to parse UUID: {e}")))?;
+    let bytes = id.as_bytes();
+    let mut u = 0u128;
+    for &b in bytes {
+        u = (u << 8) | b as u128;
```
Mcrtin, line 31: just wondering why you didn't use u128::from_le_bytes (i.e. (u128::from_le_bytes(id.as_bytes().try_into().unwrap())
(i don't know the concrete types and am prettty new to rust)
in encoded_u128_uuid_to_uuid_str too; why did you not use the u128 function for converting u128s?

src/lib/utils/config/src/statics.rs
```diff
@@ -1,21 +1,22 @@
...
 
 /// The default server configuration that is stored in memory.
 pub(crate) const DEFAULT_CONFIG: &str = include_str!("../../../../../.etc/example-config.toml");
 
 lazy_static! {
     /// The server configuration that is stored in memory.
     static ref CONFIG: ServerConfig = create_config();
+    /// The whitelist of players, wrapped in a Mutex for thread-safe access
+    /// assuming commands can come from any thread in future
+    static ref WHITELIST: Mutex<Vec<PlayerIdentity>> = Mutex::new(create_whitelist()); //should this be an rwlock instead? does it even need to be in statics?
```
ReCore-sys, line 19: Yeah should be a rwlock, but ideally you would have this as a oncelock or something, since you really only need to modify it once when the program is started

src/lib/utils/config/src/statics.rs
```diff
@@ -74,6 +75,44 @@ fn create_config() -> ServerConfig {
...
+    let mut whitelist_str = String::new();
+    if let Err(e) = file.read_to_string(&mut whitelist_str) {
+        error!("Could not read white-list file: {e}");
+        return Vec::new();
+    }
+
+    if whitelist_str.is_empty() {
+        return Vec::new();
+    }
+    serde_json::from_str::<Vec<PlayerIdentity>>(&whitelist_str).unwrap_or_else(|e| {
```
ReCore-sys, line 106: Using json seems kinda redundant. The data is not very complex, a newline/comma/space separated file format would suffice. Not a big deal tho since it should only be run once on start so performance isn't a major concern.

src/lib/core/src/identity/player_identity.rs
```diff
@@ -1,6 +1,14 @@
-#[derive(Debug)]
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
```
ReCore-sys, line 3: Serializing the player identity for whitelisting is a lot of overkill and would make it harder for server owners to manually edit the whitelist. You could just have the whitelist as a list of usernames and compare a user's name on login, having the entire PlayerIdentity require a match is going to make setting up a whitelist quite hard.

src/lib/utils/config/src/statics.rs
ReCore-sys: Actually just thought about it some more, you could just have a DashSet of usernames to check, that way you don't have to deal with any synchronization yourself and you'd have O(1) lookup. This would make using json unfeasible but i've already suggested you don't do that.

WhaleFromMars:
Should all be resolved with the latest set of commits

src/lib/core/Cargo.toml
```diff
@@ -8,3 +8,5 @@ edition = "2021"
 thiserror = { workspace = true }
 tokio = { workspace = true}
 ferrumc-ecs = {workspace = true}
+serde = { version = "1.0.215", features = ["derive"] }
```
ReCore-sys, line 11: Use workspace instead of a set version to avoid recompiling

src/lib/utils/config/src/statics.rs
```diff
@@ -1,21 +1,21 @@
...
+use uuid::Uuid;
 
 /// The default server configuration that is stored in memory.
 pub(crate) const DEFAULT_CONFIG: &str = include_str!("../../../../../.etc/example-config.toml");
 
 lazy_static! {
     /// The server configuration that is stored in memory.
     static ref CONFIG: ServerConfig = create_config();
+    /// The whitelist of player uuids.
+    static ref WHITELIST: DashMap<u128, String> = create_whitelist();
```
ReCore-sys, line 18: Use a DashSet, not a DashMap. You can either just store the uuids written in the file as a string, or you can hash the username to a u128 and store it that way. Either way storing a username and uuid for each entry is redundant and means that in this case checking a username could have up to O(n) lookup, whereas storing uuids *and* usernames as the key gets you O(1) lookup.

src/lib/utils/config/src/statics.rs
```diff
@@ -74,6 +74,94 @@ fn create_config() -> ServerConfig {
...
+
+    //read and split the file, its username:uuid format
+    let whitelist: DashMap<u128, String> = DashMap::new();
+    for (i, line) in whitelist_str.lines().enumerate() {
+        let line = line.trim();
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+
+        let mut split = line.split(':');
```
ReCore-sys, line 109: Requiring both a username and uuid can be a hassle. uuids can be matched with regex pretty easily so you can just go through each line, check if it matches a regex and if it does, its a uuid, otherwise interpret as a username. Then on login, just check if either the username or the uuid is in the whitelist, that way you can have either a player's uuid or their username and add them to the whitelist.

src/lib/net/src/packets/outgoing/login_disconnect.rs
```diff
@@ -0,0 +1,14 @@
+use ferrumc_macros::{packet, NetEncode};
+use std::io::Write;
+
+#[derive(NetEncode)]
+#[packet(packet_id = 0x00)]
```
ReCore-sys, line 5: Is this correct packet ID? It might be, I haven't checked, just seems like a coincidence.

0xnim:
why not just use minecraft format 
```json
[
  {
    "uuid": "f430dbb6-5d9a-444e-b542-e47329b2c5a0",
    "name": "username"
  },
  {
    "uuid": "e5aa0f99-2727-4a11-981f-dded8b1cd032",
    "name": "username"
  }
]
```

src/lib/utils/config/src/statics.rs
Mcrtin: > Using json seems kinda redundant. The data is not very complex, a newline/comma/space separated file format would suffice. Not a big deal tho since it should only be run once on start so performance isn't a major concern.

https://github.com/ferrumc-rs/ferrumc/pull/140#discussion_r1881363216

Mcrtin:
https://github.com/ferrumc-rs/ferrumc/pull/140#discussion_r1881363216

> Using json seems kinda redundant. The data is not very complex, a newline/comma/space separated file format would suffice. Not a big deal tho since it should only be run once on start so performance isn't a major concern

0xnim:
Using JSON is better because we can use it directly without rewriting the format. It’s a standard, so other programs can work with it easily. JSON also supports more complex structures if we need them in the future(op level for example), and there are plenty of tools available for handling it. Overall, it’s more flexible and easier to work with than a custom format like comma seperated.

0xnim:
i'm fine doing it if you dont want to

0xnim:
Also can we move this to a feature branch so i can pr to it?

Mcrtin:
If I recall correctly, it was already in json in the first place, but got changed because of that comment

WhaleFromMars:
This is for a whitelist, it was moved from vanilla file parity due to pr reviews, unless im missing something I see no need for a more complex structure in the future, I was planning on keeping the same rules as vanilla. if youre op (when implemented), you bypass it, simple as, theres no need to add a layer to configuration, I dont see why another program would need to use it, plugins will have access through whatever api gets defined in future, the goal is to make it as simple for people to use as possible, using the format in this discord message <https://discord.com/channels/1277314213878173726/1302503481432604742/1317064474963873822>

ReCore-sys:
> Using JSON is better because we can use it directly without rewriting the format. It’s a standard, so other programs can work with it easily. JSON also supports more complex structures if we need them in the future(op level for example), and there are plenty of tools available for handling it. Overall, it’s more flexible and easier to work with than a custom format like comma seperated.

I don't see why other programs need to interact with it? The reason we'd have it as as a file in the first place instead of storing in the database is so that it's easy for the end user to modify it by hand. I'm also not sure why we'd need a more complex data structure, a list of user names and uuids is fine.
Overall it's meant to be a simple format so that an average end user can just set up a quick server to play minecraft with their friends, having the whitelist be complicated to set up will deter people.

0xnim:
Server managment software uses a json as a standard. Having to write libraries just to interact with the whitelist does not sound good

0xnim:
Bud you cant just take my commits and write them yourself

0xnim:
Just merge my pr

0xnim:
This is the exact reason we have feature branches. Dony make pra directly to main

Mcrtin:
after the second hyphen, i believe there can only be an 8, 9, A, or B

GStudiosX2:
> This is the exact reason we have feature branches. Dony make pra directly to main

No one has access to make branches or push to branches other than Record and Sweatty

Skullians:
clippy is not happy

PR #143, opened 2024-12-23
command API

radstevee:
still very unfinished, basic API is in here.

TODO:
- [x] arg parsing
- [x] contexts
- [x] state
- [ ] macro builder
- [ ] derive macros

### DIFF
--- a/src/lib/commands/src/ctx.rs
+++ b/src/lib/commands/src/ctx.rs
@@ -1,0 +1,43 @@
+use std::{
+    any::Any,
+    sync::{Arc, Mutex},
+};
+
+use ferrumc_state::GlobalState;
+
+use crate::{input::CommandInput, Command};
+
+pub struct CommandContext {
+    pub input: Arc<Mutex<CommandInput>>,
+    pub command: Arc<Command>,
+    pub state: GlobalState,
+}
+
+impl CommandContext {
+    pub fn new(input: CommandInput, command: Arc<Command>, state: GlobalState) -> Arc<Self> {
+        Arc::new(Self {
+            input: Arc::new(Mutex::new(input)),
+            command,
+            state,
+        })
+    }
+
+    pub fn arg<T: Any>(&self, name: &str) -> T {
+        if let Some(arg) = self.command.args.iter().find(|a| a.name == name) {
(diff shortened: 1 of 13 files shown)


src/lib/commands/src/infrastructure.rs
```diff
@@ -0,0 +1,34 @@
+use dashmap::DashMap;
+use std::sync::{Arc, LazyLock};
+
+use crate::Command;
+
+static COMMANDS: LazyLock<DashMap<&'static str, Arc<Command>>> = LazyLock::new(DashMap::new);
```
radstevee, line 6: I'm not really a fan of the fact I have to use arcs here tbh

src/lib/commands/src/ctx.rs
```diff
@@ -0,0 +1,36 @@
...
+/// Context of the execution of a command.
+pub struct CommandContext {
+    /// The command input.
+    pub input: CommandInput,
+
+    /// The command.
+    pub command: Arc<Command>,
+
+    /// The sender of the command.
+    pub sender: Sender,
```
ReCore-sys, line 23: Should be optional since the console could send the command

src/lib/default_commands/src/nested.rs
ReCore-sys, line 1: Should default commands not be in the normal commands crate?

src/lib/net/crates/codec/src/decode/primitives.rs
```diff
@@ -145,6 +145,38 @@ where
     }
 }
 
+/// This implementation assumes that the optional was written using PacketByteBuf#writeNullable and has a leading bool.
+impl<T> NetDecode for Option<T>
```
ReCore-sys: look at prefixedoptional, it's basically the same thing

src/lib/commands/src/ctx.rs
evelynmeowww: That'd be `Sender::Server`

src/lib/default_commands/src/nested.rs
evelynmeowww: Not quite sure what you mean, but I did leave example / test commands in there, those should probably go once we actually add commands

PR #145, opened 2024-12-29
feat: Inventories

Outspending:
Adds Inventories into Ferrumc.

- [x] Simple Inventory Implementation
- [ ] Inventory Content Manager (Setting rows / columns / ect)
- [x] Inventory Syncing (Viewers System?)
- [ ] Inventory Types such as AnvilInventory, CraftingInventory, BeaconInventory, ect...
- [x] Better Inventory Packets (Instead of sending a packet per slot)
- [x] Make it compatible with ECS
- [x] Inventory Clicking
- [ ] Player Inventory Support
- [ ] Creative Inventory Support
- [ ] Slot Structure


src/lib/inventory/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub mod contents;
+pub mod inventory;
+pub mod slot;
+pub mod types;
+pub mod viewers;
+
+mod tests;
```
radstevee, line 7: Gate this behind `#[cfg(test)]`

Outspending: Will do 👍

PR #147, opened 2024-12-30
feat/entityanimation

teoscura:
1. Added two packets: 0x03 Outgoing and 0x36 Ingcoming, to support for basic player animations as detailed in the wiki. 
2. Slight readjustments to the broadcast.rs file found in `src/lib/net/utils/` that makes it easier to send any packets regarding one single player to the rest of the server. -made by @Sweattypalms 

## What needs to be done eventually:
- [ ] Observe behaviour when entities are visible.
- [ ] Make sure that all types of animations are implemented and properly visible according to vanilla behaviour.


src/lib/net/src/packets/outgoing/entity_animation.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(NetEncode)]
+#[packet(packet_id = 0x03)]
+pub struct EntityAnimationPacket {
+    pub eid: VarInt,
+    pub animation: u8,
+}
+
+#[derive(Event)]
+pub struct EntityAnimationEvent {
+    pub eid: VarInt,
```
Sweattypalms, line 17: Please save `Entity` (usize) instead of VarInt for easier access.

src/lib/net/src/packets/outgoing/entity_animation.rs
```diff
@@ -0,0 +1,53 @@
...
+    //      of the player doing it, but as long as we still cant see other players, this will be fine.
+
+    broadcast(
+        &event.packet,
+        &state,
+        BroadcastOptions::default().except([event.eid.val as usize]),
+    )
+        .await?;
+    Ok(event)
+}
```
Sweattypalms, line 37: Don't put event handlers in the net crate. Put them in the bin crate.

src/lib/net/src/packets/outgoing/entity_animation.rs
```diff
@@ -0,0 +1,53 @@
...
+impl EntityAnimationPacket {
+    pub fn new(eid: VarInt, animation: u8) -> Self {
+        Self { eid, animation }
+    }
+}
+
+impl EntityAnimationEvent {
+    pub fn new(eid: VarInt, animation: u8) -> Self {
+        Self {
+            eid: eid.clone(),
```
Sweattypalms, line 48: Save usize (entity id) instead of the VarInt, like mentioned before.

Sweattypalms (changes requested):

Sweattypalms (approved):

PR #149, opened 2024-12-30
Optimize the Importing of a world

0xnim:


src/lib/storage/src/lmdb.rs
```diff
@@ -159,6 +159,45 @@ impl LmdbBackend {
...
+            let keymap: HashMap<u128, &Vec<u8>> = data.iter().map(|(k, v)| (*k, v)).collect();
+
+            // Iterate through the keys in sorted order
+            let mut sorted_keys: Vec<u128> = keymap.keys().cloned().collect();
+            sorted_keys.sort();
+
+            // Iterate through the sorted keys to perform upserts
+            for key in sorted_keys {
+                // Check if the key already exists
+                if db.get(&rw_txn, &key)?.is_some() {
```
ReCore-sys, line 184: Why check but do the same thing anyway? That's just an extra IO operation

radstevee (approved):

src/lib/world/src/importing.rs
```diff
@@ -1,172 +1,221 @@
...
-        let progress_bar = Arc::new(ProgressBar::new(self.get_chunk_count(import_dir)?));
-        info!("Importing chunks from import directory...");
-        let start = std::time::Instant::now();
-        let mut task_set = JoinSet::new();
+        check_paths_validity(&import_dir)?;
+
+        let total_chunks = self.get_chunk_count(&import_dir)?;
+        let progress_style = ProgressStyle::default_bar()
+            .template("[{elapsed_precise}] {bar:40.cyan/blue} {pos:>7}/{len:7} {msg}")
+            .unwrap();
```
ReCore-sys, line 82: don't use unwrap

src/lib/world/src/db_functions.rs
```diff
@@ -129,6 +130,31 @@ pub(crate) async fn save_chunk_internal(world: &World, chunk: Chunk) -> Result<(
...
 }
 
+pub(crate) async fn save_chunk_internal_batch(
+    world: &World,
+    chunks: Vec<Chunk>,
+) -> Result<(), WorldError> {
+    // Prepare the batch data for the upsert
+    let mut batch_data = Vec::new();
+
+    for chunk in chunks.iter() {
```
ReCore-sys, line 140: Could do this as an iterator with rayon to multithread the compression

src/lib/world/src/importing.rs
```diff
@@ -1,172 +1,221 @@
...
-    fn get_chunk_count(&self, import_dir: PathBuf) -> Result<u64, WorldError> {
+    fn get_chunk_count(&self, import_dir: &Path) -> Result<u64, WorldError> {
         info!("Counting chunks in import directory...");
         let regions_dir = import_dir.join("region").read_dir()?;
         let chunk_count = AtomicU64::new(0);
+
         regions_dir
-            .into_iter()
             .par_bridge()
-            .for_each(|region_file| match region_file {
```
ReCore-sys, line 57: if you are removing these checks, make sure it still works if there isn't valid data to import and that there are clear messages if the data isn't valid

ReCore-sys (changes requested):
Logic is sound, could do with some performance tune ups

src/lib/storage/src/lmdb.rs
0xnim: because if you dont then lmdb has to do it anyway and lmdb does it slower and per each into the whole database

ReCore-sys: But you are doing the exact same function call, regardless of whether the condition fails or passes, so calling `get()` first won't make a difference.

0xnim: I tested

src/lib/world/src/importing.rs
0xnim: i have to? don't i?

ReCore-sys: no? Print out the error with a description or return it if the function returns a result

src/lib/world/src/importing.rs
0xnim: should not be a problem

src/lib/world/src/importing.rs
0xnim: this is the progress  bar dude

ReCore-sys (approved):

0xnim:
ty

PR #150, opened 2024-12-31
Block editing

ReCore-sys:
This PR implements block editing and some basic terrain generation.

## Description
This is a fairly large PR that implements a lot of different things, the main ones being the `set_block()` function as well as some simple terrain generation.
The `set_block()` method uses some bitwise manipulation to modify the block in a chunk. It has been tested relatively well and is unlikely to break under normal circumstances. It is currently unable to use the direct palette format but that shouldn't really be an issue for now.
There is also some very basic terrain gen implemented using OpenSimplex noise, but should absolutely be expanded on later.
There is certainly room for optimizations, but these were largely ignored for the sake of getting this core feature implemented as soon as possible.
One important note is that breaking blocks works for the client, but placing blocks is very half-baked. Inventories will need to be implemented for placing different blocks and currently the default placing of stone blocks with right click is exceptionally buggy, so this should be sorted out when inventories get done.

## Motivation and Context
Block editing is a fundamental part of minecraft, both for player interactions with the world and for terrain generation. This PR is vital for minecraft being at all playable.

## How has this been tested?
Tested manually and with unit tests

TICahXd

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.


src/lib/net/src/utils/broadcast.rs
```diff
@@ -70,7 +70,7 @@ fn get_all_entities(state: &GlobalState) -> HashSet<Entity> {
     state
         .universe
         .get_component_manager()
-        .get_entities_with::<ChunkReceiver>()
+        .get_entities_with::<PlayerIdentity>()
```
Sweattypalms, line 73: Keep this as a ChunkReceiver, since you don't want to send play packets to someone in login state.

ReCore-sys: There has to be a better solution that that. Also why would a player during login have a player identity?

Sweattypalms: You're more than welcome to find a better solution, but, player receives the PlayerIdentity component during login start packet. 
Oh wait, you also add ChunkReceiver during login start? Add it when login process is completed, and keep this as is. So 'everyone' would be players that could receive chunks I suppose?

src/lib/utils/general_purpose/src/data_packing/i32.rs
```diff
@@ -0,0 +1,109 @@
...
+/// * `Err(DataPackingError)` - If the size exceeds 32 bits or the offset plus size exceeds 64 bits.
+///
+/// # Errors
+///
+/// * `DataPackingError::SizeExceedsMaxSize` - If `size` is greater than 32.
+/// * `DataPackingError::NotEnoughBits` - If `offset + size` exceeds 64 bits.
+///    Reads an n-bit integer from a packed `i64`.
+pub fn read_nbit_i32(
+    word: &i64,
+    bit_size: usize,
```
0xnim, line 23: you state earlier that this must be 32 or less. so why not just u8?

src/bin/src/main.rs
```diff
@@ -63,10 +65,41 @@ async fn main() {
...
+            .collect();
+        for chunk in generated_chunks {
+            let chunk = chunk.map_err(|e| {
+                error!("Error generating chunk: {:?}", e);
+                BinaryError::Custom("Error generating chunk".to_string())
+            })?;
+            global_state.world.save_chunk(chunk).await?;
+        }
+        info!("Finished generating spawn chunks...");
+    }
```
Sweattypalms, line 102: Why not make this into a function, why just put it directly in the main function?

src/bin/src/systems/chunk_fetcher.rs
```diff
@@ -13,6 +16,31 @@ pub struct ChunkFetcher {
...
+                            properties: None,
+                        },
+                    )
+                    .unwrap();
+            }
+        }
+    }
+    new_chunk
+}
+
```
Sweattypalms, line 43: magic numbers?

src/bin/src/systems/chunk_fetcher.rs
```diff
@@ -41,27 +69,42 @@ impl System for ChunkFetcher {
                         let mut copied_chunks = HashMap::new();
                         for chunk in chunk_recv.needed_chunks.iter() {
                             let (key, chunk) = chunk;
-                            if chunk.is_none() {
+                            if let Fetching = chunk {
```
Sweattypalms, line 72: is that a capital F, and what is that even doing

src/bin/src/systems/chunk_fetcher.rs
```diff
@@ -79,7 +122,7 @@ impl System for ChunkFetcher {
                     }
                 }
             }
-            tokio::time::sleep(std::time::Duration::from_millis(5)).await;
+            tokio::time::sleep(std::time::Duration::from_millis(45)).await;
```
Sweattypalms, line 125: We need to rework the polling chunk system

src/lib/net/crates/codec/src/net_types/bitset.rs
```diff
@@ -9,7 +9,7 @@ pub struct BitSet(Vec<u64>);
 
 impl BitSet {
     pub fn new(size: usize) -> Self {
-        let num_blocks = (size + 63) / 64;
+        let num_blocks = size.div_ceil(64);
```
Sweattypalms, line 12: magic number?

src/lib/world_gen/src/biomes/plains.rs
```diff
@@ -0,0 +1,144 @@
...
+                            },
+                        )?;
+                    }
+                }
+            }
+        }
+
+        Ok(chunk)
+    }
+}
```
Sweattypalms, line 95: underscored function names? not used?
Also would be nice to make this function into smaller functions, right now, it's a huge function.

Sweattypalms (changes requested):

src/bin/src/main.rs
ReCore-sys: Originally it was a lot smaller but yeah you raise a good point

src/bin/src/systems/chunk_fetcher.rs
ReCore-sys: That's kinda just how terrain gen is. You kinda just gotta play around with numbers until it looks how you want. This function will be removed though and should use the proper terrain gen algorithms

src/bin/src/systems/chunk_fetcher.rs
ReCore-sys: It's an enum variant, and yeah its pretty messy. Will be redoing this entire thing.

src/bin/src/systems/chunk_fetcher.rs
ReCore-sys: 100%. Was going to leave it to a later PR but you are right, this is terrible.

src/lib/net/crates/codec/src/net_types/bitset.rs
ReCore-sys: It's a 64 bit unsigned int. I wouldn't call 64 a magic number in this context.

src/lib/world_gen/src/biomes/plains.rs
ReCore-sys: Underscored because they are currently not used and clippy tries to kill me if i don't underscore them. And yeah should probably break it down, but since this is just a very simple proof of concept terrain gen system, it'll need to be ripped up and replaced soon anyway.

src/bin/src/systems/chunk_fetcher.rs
Sweattypalms: yeah that's a good idea. merge this pr for now. let's fix it in a different pr. otherwise it'll get too much for a single pr

Sweattypalms (approved):

PR #151, opened 2024-12-31
feat: Inventories 

Outspending:
Adds Inventories into Ferrumc.

- [x] Simple Inventory Implementation
- [x] Inventory Syncing (Viewers System?)
- [x] Inventory Types such as AnvilInventory, CraftingInventory, BeaconInventory, ect...
- [x] Better Inventory Packets (Instead of sending a packet per slot)
- [x] Make it compatible with ECS
- [x] Inventory Clicking
- [x] Player Inventory Support
- [x] Creative Inventory Support
- [ ] Inventory Components


src/lib/inventory/src/contents.rs
```diff
@@ -0,0 +1,50 @@
+use crate::slot::Slot;
+use ferrumc_net::packets::outgoing::set_container_slot::NetworkSlot;
+use ferrumc_net_codec::net_types::length_prefixed_vec::LengthPrefixedVec;
+use std::collections::BTreeMap;
+
+#[derive(Debug, Clone)]
+pub struct InventoryContents {
+    pub contents: BTreeMap<i32, Slot>,
+    pub size: usize,
```
0xnim, line 9: does this need to be a usize?

Outspending: no, it doesn't need to be usize. Do you want me to change it to something else?

0xnim: unless it has something to do with indexing which the cpu does better on native 64 bit or 32 bit. always define it. here a u32 or even a u16 should suffice i think

Outspending: Yep I changed all the types to mostly u16 now, we chillin

Sweattypalms: technically speaking, you shouldn't need more than `u8`, if you're changing anyways. 255 is more than enough innit?

Outspending: Yeah, that's true. I prob should change it to a u8

src/bin/src/packet_handlers/login_process.rs
```diff
@@ -152,73 +146,70 @@ async fn handle_ack_finish_configuration(
...
-        send_keep_alive(entity_id, &state, &mut writer).await?;
-
-        let pos = state.universe.get_mut::<Position>(entity_id)?;
-        let mut chunk_recv = state.universe.get_mut::<ChunkReceiver>(entity_id)?;
-        chunk_recv.last_chunk = Some((pos.x as i32, pos.z as i32, String::from("overworld")));
-        chunk_recv.calculate_chunks().await;
-    }
 
-    player_info_update_packets(entity_id, &state).await?;
-    broadcast_spawn_entity_packet(entity_id, &state).await?;
```
GStudiosX2, line 215: This needs to be added back

src/bin/src/packet_handlers/login_process.rs
```diff
@@ -237,63 +228,3 @@ async fn send_keep_alive(
 
     Ok(())
 }
-
-async fn player_info_update_packets(entity_id: Entity, state: &GlobalState) -> NetResult<()> {
```
GStudiosX2, line 241: Same with this function

src/bin/src/packet_handlers/login_process.rs
```diff
@@ -237,63 +228,3 @@ async fn send_keep_alive(
...
-        writer
-            .send_packet(&packet, &NetEncodeOpts::WithLength)
-            .await?;
-        debug!("Sending player info update took: {:?}", start.elapsed());
-    }
-
-    Ok(())
-}
-
-async fn broadcast_spawn_entity_packet(entity_id: Entity, state: &GlobalState) -> NetResult<()> {
```
GStudiosX2, line 274: This too

GStudiosX2 (changes requested):

connor33341 (commented):
good code quality

PR #152, opened 2024-12-31
Actual multiplayer

Sweattypalms:
You can see other players move, you can crouch and uncrouch (metadata). If you leave it actually leaves.

## Description

Made it so players can see each other:
+Player Actions (Player Commands)
+Transform Event handler for seeing entities move
+Player Leave
+NetAngle
+Entity Metadata
+Head Rotation
+Spawn Entity
+Teleport Entity

Fixed broadcast
Remove entity on leave

## Motivation and Context

Doesn't solve shit. Just creates more problems.

## How has this been tested?

prayer to the god

i dont think you can put image in text like this maybe at a later time

## Types of changes

## Checklist:

- [ x ] My code follows the code style of this project.


Sweattypalms:
shit actually works for once

Sweattypalms:
can someone review the code please @ReCore-sys

src/bin/src/packet_handlers/login_process.rs
```diff
@@ -226,3 +237,58 @@ async fn send_keep_alive(
...
+async fn player_info_update_packets(entity_id: Entity, state: &GlobalState) -> NetResult<()> {
+    // Broadcasts a player info update packet to all players.
+    {
+        let packet = PlayerInfoUpdatePacket::new_player_join_packet(entity_id, state);
+
+        let start = Instant::now();
+        broadcast(
+            &packet,
+            state,
+            BroadcastOptions::default().except([entity_id]),
```
GStudiosX2, line 250: I'm pretty sure this needs to be sent to the joining player too

src/bin/src/packet_handlers/login_process.rs
```diff
@@ -226,3 +237,58 @@ async fn send_keep_alive(
...
+    }
+
+    Ok(())
+}
+
+async fn broadcast_spawn_entity_packet(entity_id: Entity, state: &GlobalState) -> NetResult<()> {
+    let packet = SpawnEntityPacket::player(entity_id, state)?;
+
+    let start = Instant::now();
+    broadcast(&packet, state, BroadcastOptions::default()).await?;
```
GStudiosX2, line 278: I'm pretty sure this shouldn't be sent to the joining player

GStudiosX2 (changes requested):
Nice but a few changes

src/lib/net/src/packets/outgoing/entity_metadata.rs
```diff
@@ -0,0 +1,269 @@
...
+            &self,
+            writer: &mut W,
+            opts: &NetEncodeOpts,
+        ) -> NetEncodeResult<()> {
+            self.index().encode_async(writer, opts).await
+        }
+    }
+}
+
+mod value {
```
ReCore-sys, line 113: Could we have some docs for this? Makes absolutely no sense at face value.

src/lib/net/src/packets/outgoing/entity_metadata.rs
```diff
@@ -0,0 +1,269 @@
...
+        Digging,
+        Sliding,
+        Shooting,
+        Inhaling,
+    }
+
+    impl EntityPose {
+        pub fn index(&self) -> VarInt {
+            use EntityPose::*;
+            let val = match self {
```
ReCore-sys, line 232: This seems kinda stupid. Idk if there are better options though.

src/bin/src/packet_handlers/login_process.rs
```diff
@@ -145,69 +152,73 @@ async fn handle_ack_finish_configuration(
...
+        // add components to the entity after the connection state has been set to play.
+        // to avoid wasting resources on entities that are fetching stuff like server status etc.
+        state
+            .universe
+            .add_component::<Position>(entity_id, Position::default())?
+            .add_component::<Rotation>(entity_id, Rotation::default())?
+            .add_component::<OnGround>(entity_id, OnGround::default())?
+            .add_component::<ChunkReceiver>(entity_id, ChunkReceiver::default())?;
+
+        let mut writer = state.universe.get_mut::<StreamWriter>(entity_id)?;
```
ReCore-sys, line 170: Would be better to scope the streamwriter lock so it gets freed once the packets are done sending. Also what's the commented numbers for?

ReCore-sys (commented):
Mostly fine, just a couple nitpicks

src/lib/net/src/packets/outgoing/entity_metadata.rs
Sweattypalms: If it's stupid and it works, then it is not stupid.

Sweattypalms: Unless you want me to use unsafe code with casting in memory, I don't think there is any other way to do this

src/bin/src/packet_handlers/login_process.rs
Sweattypalms: StreamWriter is already scoped, kindof. Gets dropped before `player_info_update_packets` and `broadcast_spawn_entity_packet`. Also I don't know what those numbers are either. I didn't add them, check git blame.

src/lib/net/src/packets/outgoing/entity_metadata.rs
ReCore-sys: Nah if that's that best we can do that's fine

src/bin/src/packet_handlers/login_process.rs
ReCore-sys: It's not dropped before the send keepalive thing. Its also not dropped before grabbing the chunk sender, which can cause deadlocks if we aren't careful

src/bin/src/packet_handlers/login_process.rs
Sweattypalms: This is not that packet for TabList. This is for `AddPlayer`, and I tried sending to the joining player and I couldn't move after joining.

GStudiosX2: > This is not that packet for TabList. This is for `AddPlayer`, and I tried sending to the joining player and I couldn't move after joining.

Player info update still need to be sent to the joining player if you want skins (well we need auth first anyway)?

Also for the spawn entity it shouldn't be sent for the joining player I'm pretty sure your doing that right now no?

PR #155, opened 2025-01-02
see-other-players with Proper movement packets and File logging

Sweattypalms:
Proper movement packets (change pos, change pos & rot, change rot, teleport) and File logging.
## Description

^

## Motivation and Context

It's required cuz I didn't implement the additional less bandwith taking packets before. And file logging is needed because why wouldn't we need file logging?

## How has this been tested?

uh 
👍
## Types of changes

## Checklist:

### DIFF
--- a/src/bin/src/packet_handlers/player/update_player_position.rs
+++ b/src/bin/src/packet_handlers/player/update_player_position.rs
@@ -2,10 +2,17 @@ use ferrumc_core::chunks::chunk_receiver::ChunkReceiver;
 use ferrumc_core::transform::grounded::OnGround;
 use ferrumc_core::transform::position::Position;
 use ferrumc_core::transform::rotation::Rotation;
-use ferrumc_macros::event_handler;
+use ferrumc_ecs::entities::Entity;
+use ferrumc_macros::{event_handler, NetEncode};
 use ferrumc_net::errors::NetError;
+use ferrumc_net::packets::outgoing::teleport_entity::TeleportEntityPacket;
+use ferrumc_net::packets::outgoing::update_entity_position::UpdateEntityPositionPacket;
+use ferrumc_net::packets::outgoing::update_entity_position_and_rotation::UpdateEntityPositionAndRotationPacket;
+use ferrumc_net::packets::outgoing::update_entity_rotation::UpdateEntityRotationPacket;
 use ferrumc_net::packets::packet_events::TransformEvent;
+use ferrumc_net::utils::broadcast::{broadcast, BroadcastOptions};
 use ferrumc_net::utils::ecs_helpers::EntityExt;
+use ferrumc_net::NetResult;
 use ferrumc_state::GlobalState;
 use tracing::trace;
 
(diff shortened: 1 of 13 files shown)


src/bin/src/packet_handlers/player/update_player_position.rs
```diff
@@ -45,6 +55,12 @@ async fn handle_player_move(
         let mut position = conn_id.get_mut::<Position>(&state)?;
         trace!("Got position 1 for player move");
 
+        delta_pos = Some((
```
ReCore-sys, line 58: SIMD might help here? would suggest benchmarking first tho

src/bin/src/packet_handlers/player/update_player_position.rs
```diff
@@ -64,5 +83,71 @@ async fn handle_player_move(
...
+            (None, Some(new_rot)) => BroadcastMovementPacket::UpdateEntityRotation(
+                UpdateEntityRotationPacket::new(entity_id, &new_rot, is_grounded),
+            ),
+            _ => {
+                return Ok(());
+            }
+        }
+    };
+
+    broadcast(&packet, state, BroadcastOptions::default().all()).await?;
```
ReCore-sys, line 150: Should limit this to only players that can see them

src/lib/utils/logging/src/lib.rs
```diff
@@ -11,10 +13,33 @@ pub fn init_logging(trace_level: Level) {
     //let console = console_subscriber::spawn();
     let env_filter = EnvFilter::from_default_env().add_directive(trace_level.into());
 
+    let is_verbose = trace_level > Level::INFO;
+
+    let file_layer = {
```
ReCore-sys, line 18: Could we also have a latest.txt that is overwritten every time the server boots up?

src/bin/src/packet_handlers/player/update_player_position.rs
GStudiosX2: > Should limit this to only players that can see them

Imo probably should be done in a different pr the previous one that used teleport didn't only send it in range either iirc

src/lib/utils/logging/src/lib.rs
Sweattypalms: Yeah definitely possible. I'm using tracing-appender library right now, with builtin shit. I'll figure it out by tomorrow hopefully.

src/bin/src/packet_handlers/player/update_player_position.rs
Sweattypalms: I'm waiting for chunks shit to be more stable to add a regional entity system. So we can have lookup for entities in a radius with the ECS.

src/bin/src/packet_handlers/player/update_player_position.rs
Sweattypalms: I don't think it should matter that much, and maybe a micro-optimization? SIMD, as far as I know, has overhead, so maybe not worth it.

GStudiosX2: > I don't think it should matter that much, and maybe a micro-optimization? SIMD, as far as I know, has overhead, so maybe not worth it.

Yeah SIMD only helps when you have a lot of elements

PR #158, opened 2025-01-04
Compile time (packet ids, registries) stuff

Sweattypalms:
Made packets use resource ids (`move_player_pos_rot`) instead of packet ids (`0x01`) for both INCOMING (Serverbound) and OUTGOING (Clientbound) packets.

OLD System:
`#[packet(packet_id = 0x1B, state = "play")]`

NEW System:
`#[packet(packet_id = "move_player_pos_rot", state = "play")]`

What's a resource id and how do I get it? 🤓 ☝️ 
You can get it from here, little timmy:
Step 1: Go to the [Protocol website](https://protocol.ferrumc.com)
Step 2: Look for a packet you want, e.g.:
image
Step 3: Look for resource name, in this case it'd be `cooldown`.
Done. 

Benefits: No hardcoding, version switching (ofcourse ignoring other problems like field orders, and state machine changes), is way easier. Just change the json file (generated via the datagenerator of 1.21.1 version `server.jar`), and recompile. 
This is compile time macros so it's basically equivalent to manually typing it yourself, runtime performance-wise.

Also there is a new `get_registry_entry!` macro, also compile time. It returns a specific ID from the `registries.json`, also generated by the datagenerator.
Example usage:
image

DATA GENERATOR: https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Data_Generators

I had to manually replace all ids with the resource ids, 😔😔. 

This is a lot of yap, where's my credits?


src/lib/derive_macros/src/lib.rs
```diff
@@ -54,3 +60,18 @@ pub fn bake_packet_registry(input: TokenStream) -> TokenStream {
...
+    static_loading::registry::get(input)
+}
+
+/// Get a packet entry from the packets.json file.
+/// returns protocol_id (as 0x??) of the specified packet.
+/// e.g. get_packet_entry!("play", "clientbound", "add_entity") -> 0x01
+#[proc_macro]
+pub fn get_packet_entry(input: TokenStream) -> TokenStream {
+    static_loading::packets::get(input)
+}
```
radstevee, line 77: Might wanna move this up to the packets section

0xnim (commented):
just one quick fix

0xnim (approved):
Honestly have no idea what half this does. APPROVED

PR #159, opened 2025-01-05
GitHub Actions Fix 

0xnim:
You asked for this so here you go


GStudiosX2 (approved):
I know like nothing about actions but all the checks passed so 🤷

PR #161, opened 2025-01-11
Async ECS fix

ReCore-sys:
This PR modifies the ECS to work in an async manner, instead of the old sync system. It switches out the dashmap for an asynchronous hashmap provided by [scc](https://docs.rs/scc/latest/scc/#hashmap), fixing the deadlocking issue we were having with multiple tasks interacting with the ECS at once.

## Description
This PR changes out the top layer of the ECS's map system, this being the hashmap that maps type ids to the sparse sets holding all the components and IDs. The underlying sparse set still uses a dashmap since I wasn't sure how it worked internally, but this sparse set has much, much less cross-task contention so is much less likely to deadlock. There is still a small chance and this will be addressed in a future PR.
Since the underlying data structures have changed to an async equivalent, all of the code that interacts with the ECS is now async, leading to some large changes across the codebase. Notably, iterator support for queries has been removed since async iterators don't exist. A similar api using `while let Some(item) = query.next().await {}` is available, but the query no longer supports the Iterator trait.

## Motivation and Context
We were having major deadlocking issues with using non-async locks in the ECS in an largely async application, so moving these sync components to async has eliminated the deadlocking issue and the server no longer freezes up randomly.

## How has this been tested?
Due to the nature of this problem testing can be inconclusive. Nevertheless we have tested to the best of our ability, both manually and with unit tests and are unable to cause a deadlock. All the methods previously used to trigger the deadlocks no longer work, but there is still a chance it could occur, since we aren't 100% sure that the ECS being sync was the issue. In all likelihood it was, but there is a small chance that the problem lies elsewhere and we just got lucky during testing.
I have however added a unit tests that spawns a load of tasks and threads that all attempt to modify the ECS at once, which logically should trigger the ECS deadlock.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have added tests to cover my changes.


src/lib/ecs/src/components/mod.rs
```diff
@@ -1,21 +1,65 @@
+/// How does the ECS work? Barely
+///
+/// ok but this is how it works.
+///
+/// You've got 2 layers of maps: You got a map for mapping a type to a storage mapping, which in 
+/// turn maps an entity ID to its component of that type. You can kinda think of it as a 
+/// `HashMap<TypeID, HashMap<entity_id, component>>` except with more borrow checker abuse. Also
+/// sparse sets are in there somewhere, but idfk how those work, i think they are just an optimised
+/// hashmap? Ask Sweatty.
```
radstevee, line 9: These could be module-level comments no? `//!`

ReCore-sys: Nah. Someone just using the ECS doesn't need to know how it works under the hood and I'd rather have user facing documentation explain how to use it. I mostly have that comment so anyone else digging through the code has an easier time wrapping their head around it.

src/lib/ecs/src/lib.rs
```diff
@@ -35,36 +41,206 @@ impl Universe {
...
+    /// ```
+    /// # use tokio_test;
+    /// # use ferrumc_ecs::Universe;
+    /// # tokio_test::block_on(async {
+    ///
+    /// let universe = Universe::new();
+    ///
+    /// let entity = universe.create_entity();
+    ///
+    /// # });
```
Mcrtin, line 57: does this get displayed correctly without the closing ```?

src/lib/ecs/src/components/mod.rs
Sweattypalms: I think a more accurate representation would be:
```rs
ThreadSafeAsyncHashMap<TypeID<T>, ThreadSafeAsyncSparseSet<Entity, T>>
```
(this is where the SparseSet is used)

src/lib/ecs/src/lib.rs
ReCore-sys: Yeah, but I added it anyway

PR #174, opened 2025-04-14
Fix SIMD on AVX2 enabled platforms, use enum_delegate for IncomingPacket

CordlessCoder:


CordlessCoder:
The force push was because I accidentally pushed some ICEs that were caused by a broken cargo index

CordlessCoder:
Currently this branch is 100% functional on Stable Rust 1.86

src/lib/net/Cargo.toml
```diff
@@ -27,4 +27,5 @@ rand = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
 byteorder = { workspace = true }
 ferrumc-state = { workspace = true }
-ferrumc-general-purpose = { workspace = true }
+crossbeam-channel = { workspace = true }
+enum_delegate.workspace = true
```
ReCore-sys, line 31: I mean it works, but why does enum_delegate have a different format to all the other imports?

src/lib/net/src/packets/mod.rs
```diff
@@ -4,38 +4,36 @@ pub mod incoming;
...
-    fn handle(
-        self,
-        conn_id: usize,
-        state: std::sync::Arc<ferrumc_state::ServerState>,
-    ) -> NetResult<()> {
-        self.handle_boxed(conn_id, state)
-    }
+use incoming::*;
+#[enum_delegate::implement(IncomingPacket)]
+pub enum AnyIncomingPacket {
```
ReCore-sys, line 18: A macro is already used to generate the match statement for the packet types, would it be worth moving this enum into the macro generation to avoid manually writing it out every time a new packet is added?

src/lib/utils/general_purpose/src/simd/arrays.rs
```diff
@@ -23,19 +22,10 @@ pub fn u8_slice_to_u32_be(input: &[u8]) -> Vec<u32> {
     );
 
     #[cfg(all(target_arch = "x86_64", not(target_os = "macos")))]
-    {
```
ReCore-sys, line 26: Since I'm not actually 100% sure, could we check if these conditionals are runtime or compile time? Since the function that checks for SIMD relies on a macro, I'm inclined to say they are compile time, so compiling on a SIMD enabled system will produce *almost* runnable code for non-SIMD systems and that will lead to some really fun bugs.

ReCore-sys (commented):
99% perfect, honestly just some small nitpicks

src/lib/net/src/packets/mod.rs
CordlessCoder: It's already handled by scripts/new_packet.py, but I'm open to that idea

src/lib/utils/general_purpose/src/simd/arrays.rs
CordlessCoder: the cfg's are compile-time, but the next line checks has_simd which is a runtime check

src/lib/net/Cargo.toml
CordlessCoder: because that's the one cargo add decided to use /shrug

ReCore-sys: Not really a problem, I'll switch it over in my code when I merge

PR #191, opened 2025-07-06
Implement YAML whitelist with in-memory and safe reload

0xnim:
This pull request drastically improves the whitelist system.

JSON, it now moves away from the custom THING, and now uses standard json.
The whitelist is now managed in memory with flushes to the file.
A new reload_whitelist function allows administrators to refresh the in-memory whitelist from the disk. This function includes a safety mechanism to prevent accidental data loss by detecting unsaved changes in memory before reloading, unless explicitly forced.


ReCore-sys:
@0xnim still failing formatting

src/lib/utils/config/src/whitelist.rs
```diff
@@ -1,215 +1,212 @@
...
+            error!("Invalid UUID format in whitelist: {}", uuid);
         }
-    });
+    }
+}
 
-    let mut uuids_to_convert: HashMap<Uuid, usize> = HashMap::new(); //pure uuid entries
-    let mut valid_lines = Vec::new(); //lines that dont need changes
-    let mut invalid_lines = Vec::new(); //entry isn't a valid uuid
+fn fetch_uuid_from_name(name: &str) -> Result<String, Box<dyn std::error::Error>> {
```
Snowiiii, line 89: Why you want to fetch the UUID using expensive API requests instead of just storing it inside the Whitelist file?

ReCore-sys: That's what it's doing. It's a function to fetch the UUID given a name and insert it into the whitelist file alongside the name.

ReCore-sys:
@0xnim you still working on this? Kinda just looks like you didn't update some of the code after the config changes. If you really don't wanna try fix it, lmk and i'll take a shot at it.

src/lib/utils/config/src/whitelist.rs
Snowiiii: But why not just use the Internal UUID from the player already saved?

ReCore-sys: There's no guarantee that it's already been saved anywhere, and even if it has the player's name may have changed since the last time they joined. I don't want to rely on internal and possibly outdated records.

Issue #192, opened 2025-07-06
Time Loop


0xnim:
### Is your feature request related to a problem?

Currently a tick loop exists, but its probably not the best to use it for time stuff. For example flushing something every 5 mins. A function of the server.

### Describe the solution / feature you'd like.

 I propose a time loop one that is standard, everything can use, internals and plugins alike. Preferably we can register a timer, and its run on an interval, so everything is not checking itself and counting every tick. Possibly with Bevy.

### Alternatives you've considered.

Keep using TPS loop. I'm not quite familiar with bevys features, that might be the path on how to do this.

### Additional Context

Bevy could possibly work. I have not looked into it yet.

### I have confimed that...

- [x] ... such a feature does not exist already.
- [ ] ... I ticked all the boxes without reading them
- [x] ... such a feature request has not been submitted already.

0xnim:
How about we have a time_loop, where we only look at the soonest due timer. And have functions to manage timers, register, etc. This way we have a single timer for all plugins and systems, and it wouldn't have any busy waiting except for the normal sleep.

ReCore-sys:
That could work, but how would we actually run the schedule? The main thread is running at a predefined tick rate so we can't really hijack that and I don't think this warrants having yet another background thread

0xnim:
We don't need to hijack anything or a new thread, we just slot it in as another ecs system.

ReCore-sys:
That's the issue, we can't guarantee that the tick rate is something sensible. Someone could set it to 1 tps and that'd likely break many timed systems that expect to be run faster, or they could set it to 500 tps and then polling would be a bottleneck at that point.

0xnim:
A thread blocking until the next timer is due could also work.

ReCore-sys:
This means that the tick rate of systems wouldn't impact the tick rate of other systems. You still wouldn't be able to guarantee a perfect tick rate since the server might be overloaded, but not much we can do about that.

0xnim:
Yes, this is the implementation I had mocked up. We don't even need to check the other schedules, only the one that comes next, and sleep/block until it is due.

PR #195, opened 2025-08-12
Day And Night Cycle

Outspending:
Pretty self explanatory

### DIFF
--- a/src/bin/src/systems/world_time.rs
+++ b/src/bin/src/systems/world_time.rs
@@ -4,23 +4,26 @@ use bevy_ecs::system::{Query, ResMut};
 use ferrumc_core::time::WorldTime;
 use ferrumc_net::{connection::StreamWriter, packets::outgoing::update_time::UpdateTimePacket};
 
-pub fn day_night_cycle_system(
-    mut players: Query<&StreamWriter>,
-    mut world_time: ResMut<WorldTime>,
-) {
+pub fn day_night_cycle_system(players: Query<&StreamWriter>, mut world_time: ResMut<WorldTime>) {
     world_time.world_age += 1;
-    world_time.time_of_day = (world_time.time_of_day + 1) % 24000; // Increment time of day, wrapping around at 24000 ticks
     world_time.tick_counter += 1;
+
+    world_time.time_of_day += 1;
+    if world_time.time_of_day >= 24000 {
+        world_time.time_of_day = 0;
+    }
 
     // Broadcast the updated time to all online players
     // This will send an update every second (20 ticks)
     if world_time.tick_counter.is_multiple_of(20) {
-        for player in players.iter_mut() {
-            if let Err(e) = player.send_packet(UpdateTimePacket {
-                world_age: world_time.world_age,
-                time_of_day: world_time.time_of_day,
-                time_of_day_increasing: true,
-            }) {
+        let packet = UpdateTimePacket {
+            world_age: world_time.world_age,
+            time_of_day: world_time.time_of_day,
+            time_of_day_increasing: true,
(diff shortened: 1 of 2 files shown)


0xnim:
Create `UpdateTimePacket` once and reuse — avoids reallocation for each player.

~~~rs
let packet = UpdateTimePacket { ... };
for player in players.iter() {
    player.send_packet(packet.clone())?;
}
~~~

0xnim:
`StreamWriter` isn’t mutated directly, so `iter()` is enough — avoids unnecessary mutable borrow.

~~~rs
for player in players.iter() {
    ...
}
~~~

src/bin/src/systems/world_time.rs
```diff
@@ -0,0 +1,31 @@
+use tracing::error;
+
+use bevy_ecs::system::{Query, ResMut};
+use ferrumc_core::time::WorldTime;
+use ferrumc_net::{connection::StreamWriter, packets::outgoing::update_time::UpdateTimePacket};
+
+pub fn day_night_cycle_system(players: Query<&StreamWriter>, mut world_time: ResMut<WorldTime>) {
+    world_time.world_age += 1;
```
ReCore-sys, line 8: Possibly add a way to control the day-night cycle speed via config? Could do it by having the time as a f64, adding fractional ticks then rounding to a u64 when sending to the client.

src/bin/src/systems/world_time.rs
```diff
@@ -0,0 +1,31 @@
...
+    world_time.tick_counter += 1;
+
+    world_time.time_of_day += 1;
+    if world_time.time_of_day >= 24000 {
+        world_time.time_of_day = 0;
+    }
+
+    // Broadcast the updated time to all online players
+    // This will send an update every second (20 ticks)
+    if world_time.tick_counter.is_multiple_of(20) {
```
ReCore-sys, line 18: Tick rate might not be 20 tps, for now add a last_sent field to the WorldTime resource but later we will add in the more granular ticking in https://github.com/ferrumc-rs/ferrumc/issues/192

src/lib/net/src/packets/outgoing/update_time.rs
```diff
@@ -0,0 +1,10 @@
+use ferrumc_macros::{packet, NetEncode};
+use std::io::Write;
+
+#[derive(NetEncode, Copy, Clone)]
+#[packet(packet_id = "set_time", state = "play")]
+pub struct UpdateTimePacket {
+    pub world_age: i64,
```
ReCore-sys, line 7: Don't really need signing, can just use a u64

src/lib/core/src/time.rs
```diff
@@ -0,0 +1,8 @@
+use bevy_ecs::resource::Resource;
+
+#[derive(Resource, Debug, Default)]
+pub struct WorldTime {
+    pub world_age: i64,
+    pub time_of_day: i64,
```
ReCore-sys, line 6: Why not u64? Gives us an extra 3 million years before overflowing

src/bin/src/systems/world_time.rs
```diff
@@ -0,0 +1,31 @@
...
+        let packet = UpdateTimePacket {
+            world_age: world_time.world_age,
+            time_of_day: world_time.time_of_day,
+            time_of_day_increasing: true,
+        };
+
+        for player in players.iter() {
+            if let Err(e) = player.send_packet(packet) {
```
ReCore-sys, line 26: Use the globalstate to check if a player is actually connected. Also probably use `.send_packet_ref()` instead to avoid the tiny copying overhead.

PR #205, opened 2025-08-23
feat: Player Data Storage

KingTimer12:
Currently, my proposal is to leverage the same database used for world storage (LMDB) and persist player data there. This follows the general idea from Bukkit/Spigot/Paper, where player data is stored inside the world folder.

But I'm still considering whether storing all player data inside data.lmdb is the best approach, or if a dedicated playerdata folder inside the world directory—where each player's data would be stored separately by UUID—might provide better organization and flexibility.

For now, the implementation will move forward with LMDB, but I’m open to feedback on which design would be preferable long-term.

### Todo (work in progress, will update as needed):
 - [X] Implement player data serialization/deserialization.
 - [X] Store player data in LMDB under a dedicated namespace.
 - [X] Load initial data and create the first player entry.
 - [x] Update data of player.
 - [x] Implement SQLite.
    - [x] Add SQLite support.
    - [x] Change database to SQLite.
 - [ ] Player Data
    - [x] Position
    - [ ] Rotation
    - [x] OnGround
    - [x] Health
    - [ ] Hunger
    - [ ] Saturation
    - [ ] Inventory & Ender Chest contents  
    - [ ] Abilities (flying, invulnerable, instabuild, etc.)  
    - [ ] Attributes (movement speed, attack damage, max health, etc.)  
    - [ ] Spawn point (bed spawn, forced spawn)  
    - [ ] Status effects (potion effects, fire, air, absorption, etc.)  
    - [ ] Advancements / Statistics  
    - [ ] Recipe book state  
    - [ ] Misc (score, shoulder entities, last death location, etc.)


ReCore-sys:
Hey thanks for getting started on this! My general plan for storing entities as a whole is to use sqlite (or turso depending on it's progress) to store them all in a normal sql database, as opposed to the kv format of lmdb. If you'd be comfortable starting on that that'd be great, otherwise just use lmdb for now and I can move it over later.

KingTimer12:
Ok, I'll switch to SQLite and use the idea of ​​the playerdata folder but in a SQL database (playerdata.db) 👍

src/bin/src/systems/shutdown_systems/send_shutdown_packet.rs
```diff
@@ -1,18 +1,32 @@
 use bevy_ecs::prelude::{Entity, Query, Res};
-use ferrumc_core::identity::player_identity::PlayerIdentity;
+use ferrumc_core::{data::player::PlayerData, identity::player_identity::PlayerIdentity};
 use ferrumc_net::connection::StreamWriter;
 use ferrumc_state::GlobalStateResource;
 use ferrumc_text::TextComponent;
 
 pub fn handle(
```
ReCore-sys, line 12: Shouldn't this be a separate system?

src/lib/core/src/data/player.rs
```diff
@@ -0,0 +1,51 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::transform::position::Position;
+
+// https://minecraft.fandom.com/wiki/Player.dat_format
+#[derive(
+    Serialize, Deserialize, Clone, Debug, Encode, Decode, Component, typename::TypeName, PartialEq,
+)]
+pub struct PlayerData {
+    pub pos: Position,
```
ReCore-sys: Isn't this just duplicating the Position, Rotation and OnGround components? Wouldn't this be better to just use those components normally to avoid desyncing them?

src/lib/core/src/data/player.rs
```diff
@@ -0,0 +1,51 @@
...
+        Self {
+            pos,
+            on_ground,
+            dimension: dimension.to_string(),
+            yaw,
+            pitch,
+        }
+    }
+
+    pub fn update_position(&mut self, new_position: Position) {
```
ReCore-sys: Why do we have setters here? Just make the the fields public.

src/lib/storage/src/sqlite.rs
ReCore-sys, line 1: Implementing this as a trait doesn't make much sense since it has quite a different use-case as the KV store, especially since we'd like be storing entities in here as well.

KingTimer12: I separated them because both are banks and have, in my view, common functions and, if they decided in the future to change the bank that stores the players, I think it would make things easier...

src/lib/core/src/data/player.rs
KingTimer12: Yeah, i removed.

src/bin/src/packet_handlers/play_packets/player_loaded.rs
```diff
@@ -24,28 +26,86 @@ pub fn handle(
...
+            Ok(loaded) => match loaded {
+                Some(loaded_data) => {
+                    *position =
+                        Position::new(loaded_data.pos.x, loaded_data.pos.y, loaded_data.pos.z);
+                    tracing::info!(
+                        "Loaded player state for {}: position=({}, {}, {}), dimension={}",
+                        player_identity.uuid.as_u128(),
+                        loaded_data.pos.x,
+                        loaded_data.pos.y,
+                        loaded_data.pos.z,
```
Sweattypalms: Implement Into str for cleaner logs

src/bin/src/packet_handlers/play_packets/player_loaded.rs
```diff
@@ -24,28 +26,86 @@ pub fn handle(
...
+                    if let Err(e) = state
+                        .0
+                        .world
+                        .save_player_state(player_identity.uuid.as_u128(), &PlayerData::default())
+                    // First time saving player data
+                    {
+                        tracing::error!(
+                            "Failed to save player state for {} ({}): {:?}",
+                            player_identity.username,
+                            player_identity.uuid.as_u128(),
```
Sweattypalms: implement into str for player identity for cleaner logs

src/bin/src/packet_handlers/player/player_disconnect.rs
```diff
@@ -0,0 +1,20 @@
...
+        let uuid = event.identity.uuid.as_u128();
+        let username = &event.identity.username;
+        if let Err(e) = state.0.world.save_player_state(uuid, &event.data) {
+            tracing::error!("Failed to save player state for {}: {}", username, e);
+        } else {
+            tracing::info!("Player state saved for {}", username);
+        }
+        cmd.entity(event.entity).despawn();
+    }
+}
```
Sweattypalms: from what i see, this is an event handler, why is this in the packet handler dir? is confusing if its the disconnect event received from connection killer and server kill.

src/lib/core/state/src/lib.rs
```diff
@@ -12,6 +14,7 @@ use std::time::Instant;
 pub struct ServerState {
     pub world: World,
     pub terrain_generator: WorldGenerator,
+    pub player_state: PlayerState,
```
Sweattypalms: is this necessary if we're using an ECS?

src/lib/core/state/src/player_state.rs
```diff
@@ -0,0 +1,22 @@
...
+        self.player_data.contains_key(&entity)
+    }
+
+    pub fn disconnect(&self, entity: Entity) {
+        self.player_data.remove(&entity);
+    }
+
+    pub fn connect(&self, entity: Entity, data: PlayerData) {
+        self.player_data.insert(entity, data);
+    }
```
Sweattypalms, line 21: again, the is_connected, disconnect, and connect, you could use the ECS to check if the entity exists, etc?

Sweattypalms (changes requested):
I've reviewed what I could. Not that experienced with database, @ReCore-sys would be able to review that section properly. But is SQL Injection a thing we should be worried about? I reckon not, since we have control over the database. And also traits for database seems interesting, again, not an expert so recore will have to check that out.

src/lib/core/state/src/lib.rs
KingTimer12: It's not, it's just that at first I was confused about how to start, so I tried several ways, but I ended up forgetting to take this one out xD

src/bin/src/packet_handlers/player/player_disconnect.rs
KingTimer12: Honestly, I didn't know where to put it, after all, we didn't have a folder called events and I didn't know whether to put it inside systems or packets (the reason for putting packets is because of the "head_rot" event)

src/bin/src/packet_handlers/play_packets/player_loaded.rs
KingTimer12: done

src/bin/src/packet_handlers/play_packets/player_loaded.rs
KingTimer12: done

src/bin/src/packet_handlers/player/player_disconnect.rs
Sweattypalms: head rot event is most likely being called from a packet. on the other hand the event you made has nothing to do with packets. why not make a separate module specifically for events? i reckon it'd be best?

ReCore-sys: I agree, having it here makes little sense

src/bin/src/systems/shutdown_systems/send_shutdown_packet.rs
```diff
@@ -1,19 +1,40 @@
...
 ) {
     let packet = ferrumc_net::packets::outgoing::disconnect::DisconnectPacket {
         reason: TextComponent::from("Server is shutting down"),
     };
-
-    for (entity, conn, identity) in query.iter() {
+    for (entity, conn, identity, position, on_ground, rotation) in query.iter() {
         if state.0.players.is_connected(entity) {
+            let player_disconnect = PlayerDisconnectEvent {
+                data: PlayerData::new(position, on_ground.0, "overworld", rotation),
```
ReCore-sys: Why are we including this data here? The event handler can just fetch this data if you give it the entity

src/lib/storage/src/database.rs
```diff
@@ -0,0 +1,33 @@
+use crate::errors::StorageError;
+
+pub trait Database {
```
ReCore-sys: This trait is largely not going to work when we need to start doing more complicated SQL queries and we won't be doing any dynamic dispatch shenanigans with multiple DB backends, so I'm not sure if using a trait here is the right move

src/lib/storage/src/sqlite.rs
```diff
@@ -0,0 +1,340 @@
...
+            "CREATE TABLE IF NOT EXISTS \"{}\" (key TEXT PRIMARY KEY, value JSON NOT NULL)",
+            table
+        );
+        conn.execute(&sql, [])?;
+        Ok(())
+    }
+
+    fn insert(&self, table: &str, key: Self::Key, value: Self::Value) -> Result<(), StorageError> {
+        let conn = self.open_conn()?;
+        let json_val: Value =
```
ReCore-sys: Why are we using json here? we know the fields, you can just store those as columns in the table, that kinda the whole point of SQL databases

src/lib/world/src/player_state.rs
```diff
@@ -0,0 +1,29 @@
...
+use ferrumc_storage::database::Database;
+
+use crate::{errors::PlayerDataError, World};
+
+// Table name for player state data in SQLite
+const TABLE_NAME: &str = "playerdata";
+
+impl World {
+    pub fn save_player_state(&self, key: u128, state: &PlayerData) -> Result<(), PlayerDataError> {
+        self.player_state_backend.create_table(TABLE_NAME)?;
```
ReCore-sys: You don't need to create a table each time you call, just generate the tables in the setup/init stage and assume it's there, sqlite will error out if it's not

ReCore-sys (changes requested):

src/lib/world/src/player_state.rs
KingTimer12: That makes sense! I had it called once during world initialization.

src/bin/src/systems/shutdown_systems/send_shutdown_packet.rs
KingTimer12: I didn't know that was possible... change made!

src/lib/storage/src/sqlite.rs
KingTimer12: When I created this, I didn't know what the data was and wanted to make it "dynamic"/"generic" in case there were additions or changes to the data, but in one of the tests, I realized that what I wanted didn't work with JSON, so it's better to just go back to the standard SQL approach. I'll do that later.

src/bin/src/packet_handlers/player/player_disconnect.rs
KingTimer12: Done!

PR #208, opened 2025-09-25
Reworked chunk format

ReCore-sys:
Our chunk format is kinda garbage. Not any more tho. Maybe.

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,2 +1,2 @@
 [profile.default]
-slow-timeout = { period = "15s", terminate-after = 2 }
+slow-timeout = { period = "15s", terminate-after = 3 }


src/bin/src/packet_handlers/play_packets/set_player_position.rs
```diff
@@ -46,9 +47,9 @@ pub fn handle(
             ((new_position.z * 4096.0) - (position.z * 4096.0)) as i16,
         ));
 
-        let old_chunk = (position.x as i32 >> 4, position.z as i32 >> 4);
+        let old_chunk = IVec2::new(position.x as i32 >> 4, position.z as i32 >> 4);
```
Mcrtin, line 50: i have been using the position of the chunkcolumn in the chunk that is closest to (-inf, -inf) for easier and faster conversion. also, i don't know if bitshifting on an i32 is rounding towards -inf or 0. if it rounds to 0, it is wrong (because the pos (-7, -9) and (5, 8) for example, would be rounded to (0, 0) even though they are in different chunks)

src/lib/world/src/chunk_ops.rs
```diff
@@ -0,0 +1,217 @@
...
+    ///
+    /// * `pos` - The chunk coordinates (x, z) as IVec2.
+    /// * `dimension` - The dimension of the chunk (e.g., "minecraft:overworld").
+    /// # Examples
+    /// ```rust
+    /// use ferrumc_world::chunk_format::Chunk;
+    /// use bevy_math::IVec2;
+    ///
+    /// let chunk = Chunk::new(IVec2::new(0, 0), "minecraft:overworld".to_string());
+    pub fn new(pos: IVec2, dimension: String) -> Self {
```
Mcrtin, line 20: i don't think having the dim in a chunk is usefull; if you really need the dim, you should be able to obtain it from the container of the chunks (and please make it an enum or at least make constants so you don't have to type out the strings). i also would like to have the pos in a wrapper over having it in the core of the chunk impl.

Mcrtin (commented):
some things i noticed

src/lib/world/src/chunk_ops.rs
ReCore-sys: The reason I have the dim in the chunk is so that the database layer can handle saving to correct table without needing data from other places. And the reason I have the dim as a string is so custom dimensions can be a thing, tho I guess we could have it as an enum for the vanilla dims and have a `Custom(String)` variant for custom dimensions.
Also what do you mean have the pos as a wrapper?

Mcrtin: by the pos **in** a wrapper i mean having a wrapper struct like 
```rust
struct PosChunk(ChunkPos, Chunk);
```
so you don't have to store the pos on each chunk if you don't need it

ReCore-sys: That's going to create a lot of conflict with what types functions and such will expect. And since a lot of different things will also need the chunk's pos, we'd end up passing PosChunk around most of the time anyway. Is there a specific issue with having the position in the chunk impl?

Mcrtin: its fine. does it also have store the height of the chunk? or how does it decide if it is 256 or 384 blocks high?

ReCore-sys: chunks are always -64 to 320 since 1.18 so it'll always be 384

Mcrtin: not in the nether?

ReCore-sys: Ah. We could either store it if you'd like or infer it from the dimension? Up to you what you want to do

Mcrtin: it's the same either way because of custom dims

PR #214, opened 2025-10-18
Block macro

ReCore-sys:
This PR adds a proc macro that can take in a block and it's properties to resolve to it's numeric ID at compile time.

## Description
Basically just do `block!("stone")` and it'll resolve to `BlockId(2u32)` at compile time. This is pretty useful for having static blocks without having to do `BlockData { name: "blahblah", properties: None }.to_block_id()` and then eating the runtime hashing cost. It's also more readable than just having the block's id as a hardcoded integer since it's more readable and will account for version differences.

## Motivation and Context
The `.to_block_id()` method on BlockData can't be optimized out so there's a runtime cost for defining static blocks and just having the id as a literal is unreadable and brittle. This allows for an easy way to move the human readable to numeric id resolution to compile time, since blocks are often used in hot loops.

## How has this been tested?
All existing BlockData instances have been swapped for the macro and it still compiles just fine. Ids seem to match up nicely.

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/derive_macros/src/block/mod.rs
+++ b/src/lib/derive_macros/src/block/mod.rs
@@ -6,6 +6,8 @@ use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
 use syn::{braced, Expr, Ident, Lit, LitStr, Result, Token};
 
+pub(crate) mod matches;
+
 const JSON_FILE: &[u8] = include_bytes!("../../../../../assets/data/blockstates.json");
 
 struct Input {
--- a/src/lib/derive_macros/src/lib.rs
+++ b/src/lib/derive_macros/src/lib.rs
@@ -1,5 +1,6 @@
 #![feature(proc_macro_quote)]
 
+use block::matches;
 use proc_macro::TokenStream;
 
 mod block;
@@ -121,4 +122,23 @@ pub fn build_registry_packets(input: TokenStream) -> TokenStream {
 #[proc_macro]
 pub fn block(input: TokenStream) -> TokenStream {
     block::block(input)
+}
+
+/// A macro to check if a block ID matches a given block name at compile time.
+/// Usage:
+/// ```
+/// # use ferrumc_macros::{match_block};
+/// # use ferrumc_world::block_id::BlockId;
+/// let block_id = BlockId(1);
+/// if match_block!("stone", block_id) {
+///     // do something
+/// }
+/// ```
+/// Unfortunately, due to current limitations in Rust's proc macros, you will need to import the
+/// `BlockId` struct manually.
+///
+/// The `minecraft:` namespace is optional and will be added automatically if not present.
+#[proc_macro]
+pub fn match_block(input: TokenStream) -> TokenStream {
+    matches::matches_block(input)
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,7 +146,6 @@ ahash = "0.8.12"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.145"
 serde_derive = "1.0.228"
-serde_yaml_ng = "0.10.0"
 base64 = "0.22.1"
 
 # When if bitcode's latest version has been changed, see if clippy still complains about
@@ -171,7 +170,6 @@ lazy_static = "1.5.0"
 quote = "1.0.41"
 syn = "2.0.106"
 proc-macro2 = "1.0.101"
-proc-macro-crate = "3.4.0"
 paste = "1.0.15"
 maplit = "1.0.2"
 macro_rules_attribute = "0.2.2"
--- a/src/lib/derive_macros/src/block/matches.rs
+++ b/src/lib/derive_macros/src/block/matches.rs
@@ -1,0 +1,62 @@
+use crate::block::JSON_FILE;
+use proc_macro::TokenStream;
+use quote::quote;
+use simd_json::base::{ValueAsObject, ValueAsScalar};
+use simd_json::derived::ValueObjectAccess;
+
+struct Input {
+    name: String,
+    id_var: syn::Expr,
+}
+
+impl syn::parse::Parse for Input {
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+        let name_lit: syn::LitStr = input.parse()?;
+        let _comma: syn::Token![,] = input.parse()?;
+        let id_var: syn::Expr = input.parse()?;
+        Ok(Input {
+            name: name_lit.value(),
+            id_var,
+        })
+    }
+}
+
+// match_block!("stone", block_id); -> "if block_id == BlockId(1) { ... }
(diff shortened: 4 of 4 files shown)


src/lib/derive_macros/src/lib.rs
```diff
@@ -96,3 +97,24 @@ pub fn get_registry_entry(input: TokenStream) -> TokenStream {
...
+/// A macro to lookup block IDs at compile time.
+///
+/// Feed in the block name as a string literal, and an optional set of properties as a map.
+/// It will output a `BlockId` struct with the correct ID for that block and properties.
+/// Usage:
+/// ```ignore
+/// let block_id = block!("stone");
+/// let another_block_id = block!("minecraft:grass_block", {snowy: true});
+/// ```
+/// Unfortunately, due to current limitations in Rust's proc macros, you will need to import the
```
evelynmeowww, line 115: from what I can see at a glance, you're just not using the fully qualified names? or is there any other reason?

Mcrtin: also if, for some reason, we can't fully qalify names, can't we wrap it in a declaritive macro?

ReCore-sys: You can fully qualify the names if you want, it'll work just the same. I haven't included it in the examples since I don't see much point but I'll add if you'd prefer.

evelynmeowww: it just means that you don't need to import BlockId yourself, generally that's a bit nicer to the user

src/lib/derive_macros/src/block/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+                        lit_int.base10_digits().to_string()
+                    } else {
+                        return syn::Error::new_spanned(
+                            &kv.value,
+                            "only string, bool, and int literals are supported as property values",
+                        )
+                        .to_compile_error()
+                        .into();
+                    }
+                }
```
lexeyOK: this can be rewritten as a match statement more cleanly, given you import syn::Lit
```suggestion
               Expr::Lit(expr_lit) => match &expr_lit.lit {
                  Lit::Str(v) => v.value(),
                  Lit::Bool(v) => v.value.to_string(),
                  Lit::Int(v) => v.base10_digits().to_string(),
                  _ => return syn::Error::new_spanned(
                      &kv.value,
                      "only string, bool, and int literals are supported as property values",
                  )
                  .to_compile_error()
                  .into(),
               },
```

src/lib/derive_macros/src/lib.rs
ReCore-sys: Oh right you meant the import path. It's actually a lot harder than it seems cos you'd need to to work from within the `ferrumc_world` crate (where `use ferrumc_world::block_id::BlockId` isn't valid) and outside it (where `crate::block_id::BlockId` isn't). The only way I could make it work was having to pass in either `crate` or `ferrumc_world` as an argument to the macro and just having to import `BlockId` explicitly felt cleaner even though it's not perfect.

src/lib/derive_macros/src/lib.rs
```diff
@@ -96,3 +97,24 @@ pub fn get_registry_entry(input: TokenStream) -> TokenStream {
...
+
+/// A macro to lookup block IDs at compile time.
+///
+/// Feed in the block name as a string literal, and an optional set of properties as a map.
+/// It will output a `BlockId` struct with the correct ID for that block and properties.
+/// Usage:
+/// ```ignore
+/// let block_id = block!("stone");
+/// let another_block_id = block!("minecraft:grass_block", {snowy: true});
+/// ```
```
lexeyOK: you will need to add `ferrumc-world` to `[dev-dependencies]` for this to compile
```suggestion
/// ```
/// # use ferrumc_world::block_id::BlockId;
/// # use ferrumc_macros::block;
/// let block_id = block!("stone");
/// let another_block_id = block!("minecraft:grass_block", {snowy: true});
/// assert_eq!(block_id, BlockId(1));
/// assert_eq!(another_block_id, BlockId(8));
/// ```
```

src/lib/derive_macros/src/block/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+            Expr::Lit(expr_lit) => {
+                if let syn::Lit::Str(lit_str) = &expr_lit.lit {
+                    lit_str.value()
+                } else if let syn::Lit::Bool(lit_bool) = &expr_lit.lit {
+                    lit_bool.value.to_string()
+                } else if let syn::Lit::Int(lit_int) = &expr_lit.lit {
+                    lit_int.base10_digits().to_string()
+                } else {
+                    "unsupported".to_string()
+                }
```
lexeyOK: same as before
```suggestion
                match &expr_lint.lit {
                    Lit::Str(v) => v.value(),
                    Lit::Bool(v) => v.value.to_string(),
                    Lit::Int(v) => v.base10_digits().to_string(),
                    _ => "unsupported".to_string(),
                }
```

src/lib/derive_macros/src/lib.rs
evelynmeowww: I remember having these issues too a while ago and talked to sweatty about it, I don't remember how I solved it though

src/lib/derive_macros/src/block/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+        .iter()
+        .filter(|v| {
+            let name_in_json = v.1.get("name");
+            if let Some(resolved_name) = name_in_json {
+                resolved_name.as_str() == Some(&name_str)
+            } else {
+                false
+            }
+        })
+        .map(|v| (v.0.parse::<u32>().unwrap(), v.1))
```
lexeyOK: this simplifies to:
```suggestion
        .filter(|(_, v)| v.get("name").as_str() == Some(&name_str))
        .map(|(k, v)| (k.parse::<u32>().unwrap(), v))
```

src/lib/derive_macros/src/lib.rs
ReCore-sys: One way to solve it would be to use a second declarative macro to fill in the path but we wouldn't be able to export that out the macros crate since that crate has to be exclusively derive or proc macros. This would lead to us either having to make another declarative macros crate or have all our macros come from one crate, except for one. In my opinion it's better in terms of ergonomics to require manual imports than deal with that.
That being said, if we do ever end up needing a crate for declarative macros i'd be happy to move it over then.

src/lib/derive_macros/src/block/mod.rs
ReCore-sys: I'm not actually sure what I was thinking with this one, thanks for catching it. I'm going to assume most of these changes you've suggest are implemented in the gist you gave me on discord

lexeyOK: > I'm not actually sure what I was thinking with this one, thanks for catching it. I'm going to assume most of these changes you've suggest are implemented in the gist you gave me on discord

yeah I did use that and others

Issue #212, opened 2025-10-16
Disconnections not working correctly


ReCore-sys:
### Describe the Bug

When leaving the server, the server still keeps the connection live for some reason but all packet sends fails. We should look into this since it fills up the terminal with errors and seems to not allow players to reconnect on some systems.

### To Reproduce

Join the server, leave, then attempt to rejoin without closing either the server or the client

### Expected Behaviour

I feel that it is not unreasonable to be able to join the server more than once without it exploding

### Screenshots / Videos

None. Why does this form require them??

### FerrumC Version

0.1.0

### Rust Version

1.90 nightly

### Server Logs

```
2025-10-04T14:31:08.533840Z DEBUG src/lib/net/src/server.rs:21: Server listening on 0.0.0.0:25565
2025-10-04T14:31:26.239790Z DEBUG src/bin/src/game_loop.rs:233: Got TCP connection from 127.0.0.1:46748
2025-10-04T14:31:26.353324Z ERROR conn{addy=127.0.0.1:46748}: src/lib/net/src/connection.rs:218: Unhandled handshake error: Packet error: Un
expected Packet: expected 0X00, received 0X0C in state Play
2025-10-04T14:31:29.743853Z DEBUG src/bin/src/game_loop.rs:233: Got TCP connection from 127.0.0.1:45844
2025-10-04T14:31:30.842804Z DEBUG src/bin/src/game_loop.rs:233: Got TCP connection from 127.0.0.1:45860
2025-10-04T14:31:30.948576Z ERROR conn{addy=127.0.0.1:45860}: src/lib/net/src/connection.rs:218: Unhandled handshake error: Packet error: Un
expected Packet: expected 0X00, received 0X0C in state Play
```
for mcrtin

### Additional Context

For some people (me) it just throws some errors in the console but nothing more than that. For some people such as @Mcrtin it won't let the player rejoin at all. This might be due to some mods since I normally run a performance modpack but i know mcrtin uses vanilla

### Is there an existing issue for this?

- [x] I have searched current existing issues before creating this.
- [ ] I checked all the boxes without reading them
- [x] I have provided all necessary and relevant details, including logs, reproduction steps, etc.

Sunnickel:
This bug extends that when leaving that sometimes you can actually rejoin, what end in the server thinking you are another player. And add you just to the list of players.

[...]

PR #219, opened 2025-10-19
fix: disconnect bug

Sunnickel:
Fixed the bug that when a player was connected to the server and then left, that the server still thought that the user was on the server. This also includes for when multiple people were on the server and one disconnected, either the server crashed, tried spamming a closed connection or added new players even though the old ones were not removed.

### set_player_position
I changed here the way its checked if the connection is still valid. Before it wouldve been checked later which would've resulted in no player removal and the try to connect to the client that closed the connection.

### head_rot
Here I added a way to handle the error if the server still tries to contact the client in a race-condition. Now it just skips the player if he cant be returned from the query properly.

### connection_killer
Here I checked when the connection gets severed. With the other changes the previous all players would've been kicked if they were 

## Motivation and Context
It was an open issue and uh staying on a server after someone left and being able to rejoin is important i think.

https://github.com/ferrumc-rs/ferrumc/issues/212

## How has this been tested?

-

## Types of changes

## Checklist:

### DIFF
--- a/src/bin/src/systems/connection_killer.rs
+++ b/src/bin/src/systems/connection_killer.rs
@@ -3,7 +3,7 @@ use ferrumc_core::identity::player_identity::PlayerIdentity;
 use ferrumc_net::connection::StreamWriter;
 use ferrumc_state::GlobalStateResource;
 use ferrumc_text::TextComponent;
-use tracing::{debug, info, trace, warn};
+use tracing::{info, trace, warn};
 
 pub fn connection_killer(
     query: Query<(Entity, &StreamWriter, &PlayerIdentity)>,


src/bin/src/packet_handlers/player/head_rot.rs
```diff
@@ -16,7 +16,17 @@ pub fn handle_player_move(
     for event in events.read() {
         let entity = event.entity;
 
-        let (rot, identity) = query.get(entity).unwrap();
+        let rot;
```
lexeyOK: use let-else here 

```suggestion
        let Ok((rot, identity)) = query.get(entity) else {
            continue;
        }
```

PR #220, opened 2025-10-19
feature: Added Join and Leave messages

Sunnickel:
Added that if a player joins the server, all other players get the standard notification in the chat and when a player leaves the same.

### Player join
sends the message to the user given

### Player leave
sends the message to the user given

### Connection-Killer
handles who the message should reach and who is disconnecting

### New Connection
Added that if the player joins that it gets broadcasted to other players.

## Motivation and Context
Was a task in project and it seemed like a good start to get into this project. Requires fixing disconnects tho since the leave message wont work otherwise.

https://github.com/orgs/ferrumc-rs/projects/3?pane=issue&itemId=134061143

## How has this been tested?

image

## Types of changes

## Checklist:

### DIFF
--- a/assets/data/configs/main-config.toml
+++ b/assets/data/configs/main-config.toml
@@ -40,6 +40,3 @@ map_size = 1_000
 cache_ttl = 60
 # How big the cache can be in kb.
 cache_capacity = 20_000
-
-whitelist = false
-
--- a/src/bin/src/packet_handlers/play_packets/set_player_position.rs
+++ b/src/bin/src/packet_handlers/play_packets/set_player_position.rs
@@ -2,6 +2,7 @@ use bevy_ecs::prelude::{Entity, EventWriter, Query, Res};
 use ferrumc_core::chunks::cross_chunk_boundary_event::CrossChunkBoundaryEvent;
 use ferrumc_core::identity::player_identity::PlayerIdentity;
 use ferrumc_net::SetPlayerPositionPacketReceiver;
+use std::sync::atomic::Ordering;
 use tracing::{debug, error, trace, warn};
 
 use crate::errors::BinaryError;
@@ -151,7 +152,7 @@ fn update_pos_for_all(
     };
 
     for (entity, conn) in conn_query.iter() {
-        if !state.players.is_connected(entity) {
+        if !state.players.is_connected(entity) || !conn.running.load(Ordering::Relaxed) {
             warn!(
                 "Player {} is not connected, skipping position update",
                 entity
@@ -161,6 +162,9 @@ fn update_pos_for_all(
                 let v = player_l.value();
                 debug!("Player list: {} - {}", k, v.1);
             }
+            state
+                .players
+                .disconnect(entity, Some(String::from("Player not connected anymore.")));
             continue;
         }
         conn.send_packet_ref(&packet)?;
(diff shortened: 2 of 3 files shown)


src/bin/src/systems/connection_killer.rs
```diff
@@ -11,6 +12,9 @@ pub fn connection_killer(
     state: Res<GlobalStateResource>,
 ) {
     while let Some((disconnecting_entity, reason)) = state.0.players.disconnection_queue.pop() {
+        let disconnecting_player: &(Entity, &StreamWriter, &PlayerIdentity) =
```
ReCore-sys: Bevy's ordering is a little weird so the disconnecting player might already be removed from the list. Can you swap out the .unwrap() for just printing and error and continuing?

src/bin/src/systems/new_connections.rs
```diff
@@ -55,9 +56,21 @@ pub fn accept_new_connections(
...
-                new_connection.player_identity.username,
+                new_connection.player_identity.username.clone(),
             ),
         );
+
+        for player in &state.0.players.player_list {
+            if player
+                .value()
+                .0
+                .eq(&new_connection.player_identity.uuid.as_u128())
```
ReCore-sys: Could you use do `if player.value().0 != new_connection... { system_messages::player_join::handle(...) }` instead of using the .eq() and continue? Makes it a little more readable.

ReCore-sys (changes requested):
Really solid work for a first contribution, with a couple tiny changes and fixes for clippy it should be good to go.

src/bin/src/systems/connection_killer.rs
Sunnickel: true didn't think of that, did it that way that it only gets it once now and afterwards just uses the object from before and skips the getting from the queue.

src/bin/src/systems/new_connections.rs
Sunnickel: Didn't think of this either, thought more into the direction of java and was too much focused on using eq. Makes almost no sense in this case.

src/bin/src/systems/connection_killer.rs
```diff
@@ -11,9 +11,21 @@ pub fn connection_killer(
     mut cmd: Commands,
     state: Res<GlobalStateResource>,
 ) {
+    let mut disconnecting_player: Option<(Entity, &StreamWriter, &PlayerIdentity)> = None;
     while let Some((disconnecting_entity, reason)) = state.0.players.disconnection_queue.pop() {
-        let disconnecting_player: &(Entity, &StreamWriter, &PlayerIdentity) =
```
ReCore-sys: Nah you kinda had the right idea originally, but this is falling more into the java/C way of doing things. I'd suggest the following
```rust
if let Some((eid, conn, identity)) = &query.get(disconnecting_entity) {
	// we have the entity here, do the main logic
} else {
	warn!("Player's entity has already been removed");
}
```

Sunnickel: true, my way would've only removed the first player in the queue and afterwards it would've just tried to disconnect the same player, since my way didn't allow an overwrite of the `disconnecting_player` variable.

Sunnickel: Not resolved tho, cause i added a new error apparently

ReCore-sys (approved):

ReCore-sys:
@Sunnickel all done? I'll merge it if you are ready.

PR #223, opened 2025-11-01
Fix: Disconnecting

Sunnickel:
Looked into the bug again in [this issue](https://github.com/ferrumc-rs/ferrumc/issues/212) and my older [PR](https://github.com/ferrumc-rs/ferrumc/pull/219). I noticed that the bug wasn't completely fixed. While I fixed some parts, others were completely ignored somehow. Probably this time too, either didn't see it or didn't test for enough edge-cases.

### connection_killer
refactored for better borrowing

### new_connection
added a disconnect_handle to the entity_commands

### connection
Heres where the magic happens. Added the ServerState and the Entity to the StreamWriter Struct. In new() we now handle that Entities can be removed on a writing error and after disconnecting there is a nice clean-up to ensure everything gets removed from ecs and the connection. 

Also on ReCore's wish I removed the kill() method.

## Motivation and Context
Worked on this bug before and there were still errors.

https://github.com/ferrumc-rs/ferrumc/issues/212

## How has this been tested?

-

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/net/src/conn_init/login.rs
+++ b/src/lib/net/src/conn_init/login.rs
@@ -210,21 +210,27 @@ pub(super) async fn login(
 
     // =============================================================================================
     // 13 Await client's teleport acceptance
-    let mut skel = PacketSkeleton::new(conn_read, compressed, Play).await?;
+    // The client may send other packets (like client_tick_end) before accepting the teleport,
+    // so we loop until we get the accept_teleportation packet
     let expected_id = lookup_packet!("play", "serverbound", "accept_teleportation");
-    if skel.id != expected_id {
-        return Err(NetError::Packet(PacketError::UnexpectedPacket {
-            expected: expected_id,
-            received: skel.id,
-            state: Play,
-        }));
-    }
-
-    let confirm_player_teleport =
-        crate::packets::incoming::confirm_player_teleport::ConfirmPlayerTeleport::decode(
-            &mut skel.data,
-            &NetDecodeOpts::None,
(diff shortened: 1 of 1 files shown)


Sunnickel:
https://github.com/ferrumc-rs/ferrumc/pull/223/commits/ad5af47d326407b0259f4f14e8ff32b689c8f4d0 is from @Tonguechaude

src/bin/src/systems/connection_killer.rs
```diff
@@ -12,48 +12,52 @@ pub fn connection_killer(
...
+            .map(|(_, _, identity)| identity.clone());
+
+        if disconnecting_player_identity.is_none() {
+            warn!("Player's entity has already been removed");
+            continue;
+        }
+
+        let disconnecting_player_identity = disconnecting_player_identity.unwrap();
+
+        for (entity, conn, player_identity) in query.iter() {
```
evelynmeowww, line 27: what's the point in doing this when you're already calling get earlier and only taking the identity?

Sunnickel: get gives me the entity that wants to disconnect current as a option tho, before the loop i just get the object if there is something in there.

down i get every player that is currently on the server, I need this to send the messages to players if someone disconnects.

Mcrtin:
can't reproduce on fix/fix-disconnect; this pr seems to fix the issue

ReCore-sys:
Cool, in that case I'll merge

PR #228, opened 2025-11-03
Add build and artifact upload workflow

BHare1985:
Reintroduces a GH actions job to build the project for all supported targets and upload the resulting executables as artifacts. The build job depends on formatting, security, and test jobs to ensure code quality to reduce artifact building

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -39,6 +39,7 @@ jobs:
         run: cargo audit
   build:
     name: Build and Upload Artifacts
+    if: github.ref == 'refs/heads/master'
     runs-on: ${{ matrix.os }}
     needs: [formatting_and_security, test]
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -37,6 +37,45 @@ jobs:
...
+      - name: Cache cargo registry and build
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-
```
BHare1985: This could just be `- uses: Swatinem/rust-cache@v2` but the removal of the artifacts came from https://github.com/ferrumc-rs/ferrumc/pull/133 so that was my reference and this was the preferred methodology back then and I don't have context on why it was changed

ReCore-sys: If you think rust-cache would be better, I'm happy to use it. I think it got changed because we were trying to speed up CI, it didn't work all that well and then we never got around to changing it back.

ReCore-sys:
@BHare1985 thank you for the contributions, if you are happy with it I'll merge this PR

.github/workflows/rust.yml
BHare1985: In my latest commit I switched this so the yml is less verbose but also to maintain consistency with the rest of the file

ReCore-sys:
Awesome, TY for your work. Releases would be nice but currently the project is changing a lot and we don't have much to actually release. At some point I would like to do a 0.0.1 release but that's not really on the cards right now. That being said, progress is really picking up these days so hopefully should only be a couple weeks/months away.

PR #230, opened 2025-11-03
command macro improvements

radstevee:
fix: allow for mutable bevy system args
improvement: add explicit error message for `World` arguments


ReCore-sys:
Fair enough, ill merge in that case

PR #231, opened 2025-11-04
update to bevy 0.17.2

radstevee:
this is mainly about the painful migration from events to messages

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,6 +140,7 @@ crossbeam-queue = "0.3.12"
 
 # Network
 ureq = "3.1.2"
+reqwest = { version = "0.12.24", features = ["json"] }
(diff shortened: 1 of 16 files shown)


src/lib/core/src/conn/conn_kill_event.rs
ReCore-sys, line 1: Where did this get moved/renamed to?

evelynmeowww: The module was untracked by you with c12466685e685e441d52106233fb58bedd12f9cd, running cargo fix probably just removed it

ReCore-sys: Whoops, my bad

PR #233, opened 2025-11-04
The biggest commit you've ever seen.

0xnim:
Fix block count initialization from 0 to 1


ReCore-sys:
Took me a few hours to read through but I think I understand most of it now

PR #238, opened 2025-11-15
[Feat] Gamemodes and fully functional /gamemode command and gamemode command arguments

Pookachu:


src/lib/commands/src/arg/gamemode.rs
```diff
@@ -0,0 +1,39 @@
...
+        // We're parsing a single word
+        PrimitiveArgument::word()
+    }
+
+    fn suggest(ctx: &mut CommandContext) -> Vec<Suggestion> {
+        let input = ctx.input.read_string();
+
+        ["survival", "creative", "adventure", "spectator"]
+            .into_iter()
+            .filter(move |s| s.starts_with(&input))
```
evelynmeowww: Filtering here isn't actually needed, it's done automatically

Pookachu: Fixed in new commit, [7bdb3ba](https://github.com/ferrumc-rs/ferrumc/pull/238/commits/7bdb3ba3f3354b156036175f06a638b5a791338a)

Simply removed the filtering and then removed the resulting variable.

PR #240, opened 2025-11-15
Rework/position dvec3 wrapper

dylanopen:
This PR makes the `Position` struct dereference to Bevy Math's DVec3 type, making mathematical operations on the Position way easier and more extensible.

## No breaking changes
- Unlike the last PR that I closed, **this is not a breaking change**.
- All API functionality remains - there are simply more fields and functions available for the Position struct, as it essentially acts as a DVec3 when calling DVec3 methods. This is done through deriving Deref.


dylanopen:
Ah, I forgot to run cargo fmt on the latest commit. I'll do that when I'm home.

PR #241, opened 2025-11-15
[feat] Pickblock packet and gamemode switcher (f3+f4)

Pookachu:


dylanopen:
Known issue: for some reason the 6th slot instead resolves to the helmet position. We have no idea why.
So it all works great... unless you try to place a block in your 6th (out of 9) hotbar slot.

src/bin/src/packet_handlers/play_packets/pick_item_from_block.rs
```diff
@@ -0,0 +1,157 @@
...
+        &PlayerAbilities,
+        &mut Inventory,
+        &mut Hotbar,
+        &StreamWriter,
+    )>,
+) {
+    for (packet, sender_entity) in events.0.try_iter() {
+        // 1. Get player's components
+        let (entity, identity, abilities, mut inventory, mut hotbar, writer) =
+            match player_inv_query.get_mut(sender_entity) {
```
ReCore-sys, line 30: Panic out here, if the components don't exist we are pretty fucked as is and it adds a lot of extra clutter

ReCore-sys (changes requested):
Why does my "block picking and game mode switcher" PR have nearly 70 changed files and events for eating?

PR #248, opened 2025-11-18
web dashboard for the server

Sweattypalms:
image

Currently, barely functional and has only placeholders. Need to add the relevant data to the `GlobalState` as well as allow console commands to be run from the website through websockets. Planned to use websockets through axum.
Test with `--features dashboard`

### DIFF
--- a/src/bin/Cargo.toml
+++ b/src/bin/Cargo.toml
@@ -52,9 +52,10 @@ ctrlc = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 crossbeam-channel = { workspace = true }
 uuid = { workspace = true }
+dhat = { version = "0.3", optional = true }
 
 [features]
-dhat = []
+dhat = ["dep:dhat"]
 dashboard = [
     "dep:ferrumc-dashboard",
     "ferrumc-config/dashboard"
--- a/src/lib/dashboard/src/lib.rs
+++ b/src/lib/dashboard/src/lib.rs
@@ -110,7 +110,7 @@ async fn static_handler(Path(path): Path<String>) -> Response {
         // Try with .html extension (for clean URLs like /about -> /about.html)
         .or_else(|| {
             if !path.contains('.') {
-                DASHBOARD_DIR.get_file(&format!("{}.html", path))
+                DASHBOARD_DIR.get_file(format!("{}.html", path))
             } else {
                 None
             }


Sweattypalms:
> OK, I see. That seems reasonable, but at some point it will definitely need some documentation, because it’s currently pretty hard to understand what’s going on under the hood.
> 
> What kind of contributions can be made to the dashboard?
> 
> By the way, some frameworks like Vue (it's the only I know for sure) can generate a single .html file plus separate asset files. Providing one standalone HTML file for the entire dashboard work too, although I have doubts about scalability as the dashboard grows larger (for example, embedding the favicon as base64 in the HTML is not a good idea if you have a lot of images and SVG icons, it'll give an HTML file of multiple Mo if I'm not misunderstanding it). I guess it depends on the scope.

I'd recommend you join our discord server: https://discord.gg/Npx5qy2H2h. Let's have a chat there. You can talk to not only me, but also other contributors of the FerrumC project!

ErdemGKSL:
I would recommend Astro or SvelteKit for build-time frontends. They can extract pure HTML, CSS, and JavaScript after the build is complete, and they are much more maintainable than a traditional templating system.

With SvelteKit, for example, you can use shadcn components and customize them for FerrumC. The SvelteKit syntax is very similar to standard HTML, so the learning curve should be smooth and easy.

In this context, the goal of a “single binary” is still achievable while having a frontend that is significantly easier to maintain.

On the other hand, I wouldn’t recommend Vue, since the extracted HTML files also include Vue’s client-side library itself, which effectively introduces unnecessary bloat.

Sweattypalms:
> I would recommend Astro or SvelteKit for build-time frontends. They can extract pure HTML, CSS, and JavaScript after the build is complete, and they are much more maintainable than a traditional templating system.
> 
> With SvelteKit, for example, you can use shadcn components and customize them for FerrumC. The SvelteKit syntax is very similar to standard HTML, so the learning curve should be smooth and easy.
> 
> In this context, the goal of a “single binary” is still achievable while having a frontend that is significantly easier to maintain.
> 
> On the other hand, I wouldn’t recommend Vue, since the extracted HTML files also include Vue’s client-side library itself, which effectively introduces unnecessary bloat. Meanwhile sveltekit is just a super powered html with some custom macros kind of similar to how rust works.

I, too, think having a simple templating framework (like Svelte, or HTMX) would benefit the dashboard and the cleanliness of the code, although due to my lack of knowledge of such frameworks I've currently made the dashboard use vanilla HTML and JavaScript to get momentum going, you're more than welcome to fork and provide a better solution!

PR #251, opened 2025-11-20
feat : basics components for entity in bevy ECS

Tonguechaude:
This is a rework of my fist PR based on ferrumc-data crates


src/lib/entities/src/components/physical.rs
```diff
@@ -0,0 +1,148 @@
+use bevy_ecs::prelude::Component;
+use ferrumc_data::generated::entities::EntityType as VanillaEntityType;
+
+/// Entity bounding box (collision box).
+///
+/// Represents the volume occupied by an entity in the world.
+/// Used for collision detection and physics.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct BoundingBox {
```
ReCore-sys, line 9: Could we use bevy's AABB for this?

Tonguechaude: The problem is that Aabb3d would need to be recalculated every frame whenever the entity moves. My BoundingBox stores the entity's relative dimensions (half-width, height) which are constant. They only change when the entity changes state (baby pig, crouching player, etc.), not on every position update.

PR #252, opened 2025-11-21
Feature/encryption

MSKatKing:
This branch implements encryption with the Minecraft client. During the login phase, the client will send the server a shared key, and the server uses this key to encrypt outgoing packets and decrypt incoming packets.

It works as a wrapper around the `OwnedHalfRead` and `OwnedHalfWrite` structs to use an AES/CFB8 cipher to encrypt/decrypt data.

There is also a config toggle to enable/disable encryption.

### DIFF
--- a/src/lib/net/Cargo.toml
+++ b/src/lib/net/Cargo.toml
@@ -37,7 +37,6 @@ indexmap = { workspace = true }
 lazy_static = { workspace = true }
 yazi = { workspace = true }
 ferrumc-inventories = { workspace = true }
-ureq = { workspace = true }
 base64 = { workspace = true }
 
 
--- a/src/lib/net/crates/encryption/src/lib.rs
+++ b/src/lib/net/crates/encryption/src/lib.rs
@@ -104,34 +104,35 @@ pub fn minecraft_hex_digest(server_id: &str, shared_secret: &[u8]) -> String {
     bigint.to_str_radix(16)
 }
 
-/// Minecraft's custom implementation of a SHA1 hex digest.
-///
-/// # Note
-/// This function is only to be used in tests. The only different between this function and `minecraft_hex_digest`
-/// is that this function does **NOT** hash the encryption key.
-///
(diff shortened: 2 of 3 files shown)


src/lib/net/crates/encryption/src/lib.rs
```diff
@@ -1,16 +1,151 @@
...
+/// is that this function does **NOT** hash the encryption key.
+///
+/// # Parameters
+/// - `server_id`: The server id sent to the player in an EncryptionRequest packet.
+/// - `shared_secret`: The shared secret returned by the player in an EncryptionResponse packet.
+///
+/// # Returns
+/// - `String`: The resulting hex representation of the SHA1 digest
+#[allow(unused)]
+fn minecraft_hex_digest_test(server_id: &str, shared_secret: &[u8]) -> String {
```
ReCore-sys: You can chunk this in the conditionally compiled mod as well

src/lib/net/src/errors.rs
```diff
@@ -10,6 +10,9 @@ pub enum NetError {
     #[error("Encryption Error: {0}")]
     EncryptionError(#[from] NetEncryptionError),
 
+    #[error("Authentication Error: {0}")]
+    AuthenticationError(String),
```
ReCore-sys: you should probably just use `AuthenticationError(#[from] AuthenticationError)` here

src/lib/net/Cargo.toml
```diff
@@ -37,6 +37,8 @@ indexmap = { workspace = true }
 lazy_static = { workspace = true }
 yazi = { workspace = true }
 ferrumc-inventories = { workspace = true }
+ureq = { workspace = true }
```
ReCore-sys: We won't need ureq since we are already in an async environment, we are probably just going to use reqwest

ReCore-sys (changes requested):
Looks really good, only a couple nitpicks

PR #253, opened 2025-11-21
feature/auth

MSKatKing:
Implements authentication with Mojang's servers.

### DIFF
--- a/src/lib/net/src/auth.rs
+++ b/src/lib/net/src/auth.rs
@@ -1,33 +1,11 @@
+use crate::errors::NetAuthenticationError;
 use base64::Engine;
 use ferrumc_core::identity::player_identity::PlayerProperty;
 use ferrumc_net_encryption::minecraft_hex_digest;
 use serde_derive::Deserialize;
-use std::error::Error;
 use std::str::FromStr;
 use std::sync::Arc;
-use thiserror::Error;
 use uuid::Uuid;
-
-#[derive(Debug, Clone, Error)]
-pub enum NetAuthenticationError {
-    #[error("Failed to reach Mojang's authentication servers")]
-    CouldNotReachMojang,
-
-    #[error("The server has exceeded the rate limit allowed by Mojang")]
-    RateLimitReached,
-
-    #[error("The user could not be authenticated")]
-    FailedToAuthenticate,
-
-    #[error("Could not parse auth server response: {0}")]
-    ParseError(#[from] Arc<dyn Error + Send + Sync>),
-
-    #[error("Mojang returned a corrupted UUID")]
-    CorruptUuid,
-
-    #[error("Mojang responded with status code {0}")]
-    StatusError(u16),
-}
 
 /// Authenticates the given player with Mojang's session server.
 ///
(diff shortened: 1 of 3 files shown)


src/lib/net/src/conn_init/login.rs
```diff
@@ -129,7 +132,87 @@ pub(super) async fn login(
...
+                }
+
+                let mut response = ureq::get(&url).call().map_err(|err| {
+                    NetError::AuthenticationError(format!(
+                        "Failed to connect to Mojang servers: {}",
+                        err
+                    ))
+                })?;
+
+                match response.status().as_u16() {
```
evelynmeowww: should probably add 429 since you hit rate limits really quickly

MSKatKing: Yeah, I should probably look up the major error codes and handle them here

src/lib/net/src/conn_init/login.rs
```diff
@@ -129,7 +133,87 @@ pub(super) async fn login(
...
+                }
+
+                #[derive(Deserialize, Debug)]
+                struct AuthProperty {
+                    name: String,
+                    value: String,
+                    signature: String,
+                }
+
+                let mut response = ureq::get(&url).call().map_err(|err| {
```
ReCore-sys: Can we use reqwest here instead of ureq?

src/lib/net/src/auth.rs
```diff
@@ -0,0 +1,112 @@
...
+use ferrumc_net_encryption::minecraft_hex_digest;
+use serde_derive::Deserialize;
+use std::error::Error;
+use std::str::FromStr;
+use std::sync::Arc;
+use thiserror::Error;
+use uuid::Uuid;
+
+#[derive(Debug, Clone, Error)]
+pub enum NetAuthenticationError {
```
ReCore-sys: Can we chuck this in an errors file if possible?

src/lib/net/src/conn_init/login.rs
```diff
@@ -131,7 +133,12 @@ pub(super) async fn login(
             // =============================================================================================
             // 3.1 Authenticate the player with Mojang's servers (if online_mode is enabled)
             if get_global_config().online_mode {
-                // TODO: auth code should go here
+                let (username, uuid, properties) =
+                    authenticate_user(&login_start.username, "", &shared_secret).await?;
+
+                login_start.username = username;
```
ReCore-sys: Could we add a comparison here to see if the unvalidated uuid/name given by the client matches mojang's response? Would prevent account spoofing

src/lib/net/src/auth.rs
```diff
@@ -0,0 +1,112 @@
...
+    );
+
+    let response = reqwest::get(&url)
+        .await
+        .map_err(|_| NetAuthenticationError::CouldNotReachMojang)?;
+
+    match response.status().as_u16() {
+        200 => Ok(()),
+        204 => Err(NetAuthenticationError::FailedToAuthenticate),
+        404 => Err(NetAuthenticationError::CouldNotReachMojang),
```
ReCore-sys: Probably change this to `BadUrl` or something since we can reach mojang's servers at this point

ReCore-sys (changes requested):

PR #258, opened 2025-11-21
Spawn pig command

Tonguechaude:
image

Allow player to spawn a lobotomized pig with `/spawn pig`


src/bin/src/systems/listeners/mod.rs
```diff
@@ -2,11 +2,14 @@ pub mod digging_system;
 pub mod gamemode_change;
 pub mod player_join_message;
 pub mod player_leave_message;
+pub mod spawn_pig;
```
Sweattypalms: Name it something more generic than just 'pig'.

src/bin/src/systems/listeners/spawn_pig.rs
```diff
@@ -0,0 +1,108 @@
...
+        // Get player position and rotation
+        if let Ok((pos, rot)) = query.get(request.player_entity) {
+            // Calculate spawn position 2 blocks in front of the player
+            let yaw_radians = rot.yaw.to_radians();
+            let spawn_pos = Position::new(
+                pos.x - (yaw_radians.sin() as f64 * 2.0),
+                pos.y,
+                pos.z + (yaw_radians.cos() as f64 * 2.0),
+            );
+
```
Sweattypalms: Make this a helper function in the Position/Rotation struct to get a block "infront" of the position+rotation.

src/bin/src/systems/listeners/spawn_pig.rs
```diff
@@ -0,0 +1,108 @@
...
+                    );
+
+                    // Broadcast to all connected players
+                    let mut writer_query = world.query::<&StreamWriter>();
+                    for writer in writer_query.iter(world) {
+                        if let Err(e) = writer.send_packet_ref(&spawn_packet) {
+                            error!("Failed to send spawn packet: {:?}", e);
+                        }
+                    }
+                });
```
Sweattypalms: Should be pig specific code. Make this code modular.

src/lib/core/src/identity/entity_identity.rs
```diff
@@ -0,0 +1,60 @@
+use bevy_ecs::prelude::Component;
+use std::sync::atomic::{AtomicI32, Ordering};
+
+/// Global entity ID counter for non-player entities.
+///
+/// Each spawned entity (pig, cow, arrow, etc.) needs a unique network ID
+/// to be identified in packets. This counter generates sequential IDs
+/// starting from a high number to avoid collisions with player short_uuid.
+static ENTITY_ID_COUNTER: AtomicI32 = AtomicI32::new(1_000_000);
```
Sweattypalms, line 9: If we're using bevy, would this be needed? And it feels like a workaround/bandage solution.

src/lib/core/src/identity/entity_identity.rs
```diff
@@ -0,0 +1,60 @@
...
+            uuid,
+        }
+    }
+}
+
+impl Default for EntityIdentity {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
Sweattypalms, line 60: Aren't entity "ids" specifically handled by bevy?

src/lib/default_commands/src/spawnpig.rs
```diff
@@ -0,0 +1,23 @@
...
+            sender.send_message(TextComponent::from("Pig spawned!"), false);
+        }
+        Sender::Server => {
+            sender.send_message(
+                TextComponent::from("Only players can use this command"),
+                false,
+            );
+        }
+    }
+}
```
Sweattypalms: perhaps make it /spawn <ENTITY_TYPE>, but only allow PIG for now in the autocomplete and reject others?

src/lib/entities/src/spawn_queue.rs
```diff
@@ -0,0 +1,32 @@
...
+    SPAWN_QUEUE.push(SpawnRequest {
+        entity_type,
+        player_entity,
+    });
+}
+
+/// Pop a spawn request from the queue (called from system)
+pub fn pop_spawn_request() -> Option<SpawnRequest> {
+    SPAWN_QUEUE.pop()
+}
```
Sweattypalms: Why this instead of alternatives like tokio channels or bevy events (messages)?

src/lib/net/src/packets/outgoing/spawn_entity.rs
```diff
@@ -29,6 +30,34 @@ pub struct SpawnEntityPacket {
...
+        position: &Position,
+        rotation: &Rotation,
+    ) -> Self {
+        Self {
+            entity_id: VarInt::new(entity_id),
+            entity_uuid,
+            r#type: VarInt::new(entity_type_id),
+            x: position.x,
+            y: position.y,
+            z: position.z,
```
Sweattypalms: is it possible to make a spread handler for position that exposes the x, y, z fields? so you could do:
```rs
...position
```
which would automatically assign x, y, and z properties?

src/lib/net/src/packets/outgoing/spawn_entity.rs
```diff
@@ -29,6 +30,34 @@ pub struct SpawnEntityPacket {
...
+            pitch: NetAngle::from_degrees(rotation.pitch as f64),
+            yaw: NetAngle::from_degrees(rotation.yaw as f64),
```
Sweattypalms: Similar to the above comment, automatic spreader for rotation -> pitch, yaw?

src/lib/net/src/packets/outgoing/spawn_entity.rs
```diff
@@ -54,4 +83,37 @@ impl SpawnEntityPacket {
...
+    /// * `entity_type_id` - Protocol ID for the entity type (from vanilla data)
+    /// * `query` - Query to get entity components
+    pub fn entity(
+        entity: Entity,
+        entity_type_id: u16,
+        query: Query<(&EntityIdentity, &Position, &Rotation)>,
+    ) -> Result<Self, NetError> {
+        let (identity, position, rotation) = query.get(entity).unwrap_or_else(|_| {
+            panic!("Failed to get entity identity, position, and rotation for entity: {entity:?}")
+        });
```
Sweattypalms: Why panic instead of using the `Result` return type of the function?

Sweattypalms (changes requested):
Hey, I've reviewed your PR. It's mostly improvements/nitpicks to make the code more generalized, and also some small fixes here and there.

src/lib/entities/src/spawn_queue.rs
Tonguechaude: Command doesn't have bevy system_parameter so they can't interact with the message_writer so I do something to quick test but let me try something

Tonguechaude: My theory works I was just dumb

src/lib/core/src/identity/entity_identity.rs
ReCore-sys: Nope, bevy's entity id's are only for use with bevy and a lot of things will break if we try use them. For one, they are somewhat randomized and can't be so it'd be impossible to make entities persist after restarting the server. Bevy's entities and game entities are very different.

src/lib/core/src/identity/entity_identity.rs
```diff
@@ -33,28 +39,32 @@ pub struct EntityIdentity {
...
     /// The UUID is randomly generated.
-    pub fn new() -> Self {
+    ///
+    /// # Arguments
+    ///
+    /// * `entity` - The Bevy entity this identity belongs to
+    pub fn from_entity(entity: Entity) -> Self {
         Self {
-            entity_id: ENTITY_ID_COUNTER.fetch_add(1, Ordering::SeqCst),
+            entity_id: (entity.index() as i32).saturating_add(ENTITY_ID_OFFSET),
```
ReCore-sys: Hey sorry to do this to you but @Sweattypalms was wrong. If you use bevy's IDs you are going to run into a lot off issues with both persistence and ID collisions. Bevy's ids have a specific format they use that includes stuff like prefixes and suffixes so clamping it down to an i32 is going to get you IDs that are much less unique. You'd be better off going back to an atomic counter or just generating random UUIDs

ReCore-sys (changes requested):
sorry to ask for you to revert stuff @Sweattypalms asked for but it'll lead to a lot of issues if you don't

src/lib/core/src/identity/entity_identity.rs
MSKatKing: Would generating IDs based on runtime of the server be a possibility in the future? I think that would pretty much guarantee unique IDs for every entities (players including).

Tonguechaude: Do not be sorry it's okay it's just a git revert. I understand

PR #271, opened 2025-11-24
fix : update and modernize dockerfile

Tonguechaude:


.dockerignore
```diff
@@ -0,0 +1,32 @@
+# Build artifacts
+target/
+**/target/
+
+# IDE /!\ F*** use neovim !!
```
ReCore-sys: ???

Dockerfile
```diff
@@ -1,29 +1,43 @@
...
 WORKDIR /app
-RUN apk add --no-cache wget
-COPY --from=build-stage /app/target/release/ferrumc /app/
+RUN addgroup -S ferrumc && adduser -S ferrumc -G ferrumc
+COPY --from=builder /app/target/release/ferrumc /app/
+RUN chown -R ferrumc:ferrumc /app
 
 # leaving this out for now since we dont have world loading
 # Once we get the website working, we should put the file on our own server, instead of a contributors side project.
-# RUN wget --progress=dot:giga https://aboleth.ai/static/ferumc-121-exampleworld.tar.gz -P /app
```
ReCore-sys, line 21: Probably remove this, this was before we had world gen and the world won't wouldn't work anyway. Also I think the url is dead too.

README.md
```diff
@@ -187,6 +187,12 @@ Options:
             - `warn` (Only warnings)
             - `error` (Only errors)
 
+### [Option 3] Use Docker
+
+1. You need to build the image `sudo docker buildx build -t ferrumc:yourtag --load .`
```
ReCore-sys: Just leave it with normal docker build, if people have specific workflow setups they should know how to adapt the command.

ReCore-sys (changes requested):

.dockerignore
Sweattypalms: LMAO

Tonguechaude: let me do this please, I need to do this propaganda

PR #272, opened 2025-11-25
feature/crafting

MSKatKing:
Implements crafting with the survival inventory crafting grid. This does not implement a crafting table as those UIs have not been implemented yet. However, functionality to get a recipe from a 3x3 crafting grid does exist (`ferrumc_core::crafting::get_recipes_from_3x3`) so crafting tables should be fairly simple to implement once we get there.

When the survival crafting grid is updated, the code will search through recipes to find ones that match the current grid. If one is found, it will update the output to reflect that.

To make this work I had to implement basic inventory syncing. My code just assumes that the client is telling the truth and updates the player's inventory in whatever way the client tells it to. While for now this is fine, when we implement other containers (or even just the player inventory better) we want to ensure that changes are valid. Additionally, the crafting grid just gets cleared when the player closes their inventory. Again, this should be changed in later implementations of the inventory.

This PR does **NOT** add recipe searching functions for any other crafting type (smelting, blasting, campfire cooking, smithing, etc). It only implements the player's survival crafting grid (with the aforementioned functionality to support 3x3 grids).

If this PR gets merged I'm going to open issues for the inventory syncing problem so if someone else wants to tackle it, they can.


src/bin/src/packet_handlers/play_packets/close_container.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub fn handle(receiver: Res<CloseContainerReceiver>, mut inventories: Query<&mut Inventory>) {
+    for (event, eid) in receiver.0.try_iter() {
+        // 0 is the player's inventory
+        if event.window_id.0 == 0 {
+            debug!("Clearing crafting grid");
+
+            if let Ok(mut inventory) = inventories.get_mut(eid) {
+                for slot_id in
+                    defined_slots::player::CRAFT_SLOT_OUTPUT..=defined_slots::player::CRAFT_SLOT_4
```
Sweattypalms, line 20: Could you add some documentation as to what this means? And why it is looping through supposed 'OUTPUT' to '4' slots?

src/bin/src/packet_handlers/play_packets/close_container.rs
```diff
@@ -0,0 +1,25 @@
...
+                {
+                    inventory
+                        .set_item_with_update(slot_id as _, InventorySlot::empty(), eid)
+                        .expect("failed to write to inventory");
```
Sweattypalms: gracefully handle error

src/bin/src/packet_handlers/player/update_crafting.rs
```diff
@@ -0,0 +1,54 @@
...
+    let craft_slot_3 = inventory
+        .get_item(defined_slots::player::CRAFT_SLOT_3 as _)
+        .expect("failed to get craft slot 3")
+        .and_then(|slot| slot.item_id)
+        .and_then(|id| Item::from_id(id.0 .0 as _));
+    let craft_slot_4 = inventory
+        .get_item(defined_slots::player::CRAFT_SLOT_4 as _)
+        .expect("failed to get craft slot 4")
+        .and_then(|slot| slot.item_id)
+        .and_then(|id| Item::from_id(id.0 .0 as _));
```
Sweattypalms: Make a function that does takes in an inventory ref and inventory slot, and returns an item perhaps?

src/bin/src/packet_handlers/player/update_crafting.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        let slot = InventorySlot {
+            item_id: Some(ItemID(VarInt(item.id as _))),
+            count: VarInt(first.count as _),
+            ..Default::default()
+        };
+
+        inventory
+            .set_item_with_update(defined_slots::player::CRAFT_SLOT_OUTPUT as _, slot, eid)
+            .expect("failed to set item in inventory");
```
Sweattypalms: gracefully handle

src/bin/src/packet_handlers/player/update_crafting.rs
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+        inventory
+            .clear_slot_with_update(defined_slots::player::CRAFT_SLOT_OUTPUT as _, eid)
+            .expect("failed to clear item in inventory");
```
Sweattypalms: gracefully handle

src/lib/core/src/crafting.rs
```diff
@@ -0,0 +1,229 @@
...
+    for allowed in symbol_allowed {
+        if let Some(tag) = allowed.strip_prefix('#') {
+            if let Some(values) = TagData::get_item_tag(tag) {
+                if values.values.contains(&item_id) {
+                    return true;
+                }
+            }
+        } else if *allowed == item_id {
+            return true;
+        }
```
Sweattypalms: very deep nesting of if statements

src/bin/src/packet_handlers/play_packets/close_container.rs
MSKatKing: The OUTPUT slot has id 0 and the input slots are ids 1-4, I'll make this more clear in the code :)

PR #274, opened 2025-11-27
feature/brand

MSKatKing:
Sends a server brand string to the client during configuration. Doesn't do anything functionally, just displays it in the F3 menu and in logs. Brand string is in this format: "FerrumC v{version}" and if the build is a debug build " DEBUG" is added to the end of that string.

### DIFF
--- a/src/lib/net/src/packets/outgoing/client_bound_plugin_message.rs
+++ b/src/lib/net/src/packets/outgoing/client_bound_plugin_message.rs
@@ -1,23 +1,33 @@
 use ferrumc_macros::{packet, NetEncode};
-use ferrumc_net_codec::encode::{NetEncode, NetEncodeOpts};
+use ferrumc_net_codec::encode::NetEncode;
 
 const SERVER_VERSION: &str = concat!("FerrumC v", env!("FERRUMC_VERSION"), env!("BUILD_TYPE"));
 
+type BrandChannelData = String;
+
 #[derive(NetEncode, Clone)]
 #[packet(packet_id = "custom_payload", state = "configuration")]
-pub struct ClientBoundPluginMessagePacket {
+pub struct ClientBoundPluginMessagePacket<T: NetEncode> {
     pub channel: String,
-    pub data: Vec<u8>,
+    pub data: T,
 }
 
-impl ClientBoundPluginMessagePacket {
-    pub fn brand() -> ClientBoundPluginMessagePacket {
-        let mut data = Vec::<u8>::new();
-        SERVER_VERSION
-            .encode(&mut data, &NetEncodeOpts::None)
-            .expect("failed to write brand name to buffer");
+impl ClientBoundPluginMessagePacket<BrandChannelData> {
+    pub fn brand() -> ClientBoundPluginMessagePacket<BrandChannelData> {
         Self {
             channel: "minecraft:brand".to_string(),
+            data: SERVER_VERSION.to_string(),
+        }
+    }
+}
+
+impl<T: NetEncode> ClientBoundPluginMessagePacket<T> {
+    pub fn with_data_channel(
+        channel: impl Into<String>,
+        data: T,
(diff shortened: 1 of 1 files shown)


src/lib/net/src/packets/outgoing/client_bound_plugin_message.rs
```diff
@@ -0,0 +1,24 @@
+use ferrumc_macros::{packet, NetEncode};
+use ferrumc_net_codec::encode::{NetEncode, NetEncodeOpts};
+
+const SERVER_VERSION: &str = concat!("FerrumC v", env!("FERRUMC_VERSION"), env!("BUILD_TYPE"));
+
+#[derive(NetEncode, Clone)]
+#[packet(packet_id = "custom_payload", state = "configuration")]
+pub struct ClientBoundPluginMessagePacket {
+    pub channel: String,
+    pub data: Vec<u8>,
```
Sweattypalms: Why not just make it a string and let the NetEncode macro automatically serialized it? I feel like that's what youre doing anyways right?

MSKatKing: Currently that's all it's being used for, but this packet it meant to be used for a bunch of different things with different data formats. The `minecraft:brand` channel just only needs a String. Especially for the future (even more so if we support plugins) I feel it's better to just keep it generic so we don't need to keep coming back and changing it.

MSKatKing: If you wanted I suppose I could turn it into an enum with explicit types for the minecraft channels and then a generic one for anything else.

Sweattypalms: Functionality-wise it seems fine, I was just wondering if it was necessary to manually encode the data in the constructor, but I suppose it's fine since you say there can be multiple types. Although you did mention the idea of turning it into an Enum with explicit types, I suppose that'd be the best? I believe the NetEncode derive supports enums.

src/lib/net/src/packets/outgoing/client_bound_plugin_message.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub fn brand() -> ClientBoundPluginMessagePacket {
+        let mut data = Vec::<u8>::new();
+        SERVER_VERSION
+            .encode(&mut data, &NetEncodeOpts::None)
+            .expect("failed to write brand name to buffer");
+        Self {
+            channel: "minecraft:brand".to_string(),
+            data,
+        }
+    }
```
Sweattypalms: encoding this everytime this function is called? is it possible to make the encoding const or compile time? it's going to be same everytime regardless? encoding everytime seems redundant.

ReCore-sys: We could look into making encode a const fn, but I don't reckon that's within the scope of this PR

src/lib/net/src/packets/outgoing/client_bound_plugin_message.rs
MSKatKing: I thought about it a bit more and decided that I think it would be better if I changed `data` to a generic type that implements NetEncode. I think that this is the most extensible, and also utilizes the macro to do encoding for me.

src/lib/net/src/packets/outgoing/client_bound_plugin_message.rs
MSKatKing: I changed the code so that it uses a generic type instead, and for the brand channel it uses a String as the type. This removes the need to encode a String twice (one into the data Vec and another from the Vec into the packet).

src/lib/net/src/packets/outgoing/client_bound_plugin_message.rs
Sweattypalms: that's way better, i love that. thank you;

PR #283, opened 2025-11-30
Fix/cleanup

Sweattypalms:
cleanup some code and refactor it.

### DIFF
--- a/docs/protocol-flow.md
+++ b/docs/protocol-flow.md
@@ -229,7 +229,8 @@ flowchart TD
 │  └─────────────────────────────────────────────────────┘    │
 └─────────────────────────────────────────────────────────────┘
                               │
-                              │ crossbeam channels
+                              │ crossbeam channels (incoming)
+                              │ tokio mpsc channels (outgoing)
                               ▼
 ┌─────────────────────────────────────────────────────────────┐
 │                  NETWORK THREAD (Tokio)                     │


docs/protocol-flow.md
```diff
@@ -0,0 +1,256 @@
...
+│                     MAIN THREAD                             │
+│  ┌─────────────────────────────────────────────────────┐    │
+│  │              Bevy ECS Scheduler                     │    │
+│  │  - Tick schedule (20 TPS)                           │    │
+│  │  - Keepalive schedule (1s)                          │    │
+│  │  - World sync schedule (15s)                        │    │
+│  └─────────────────────────────────────────────────────┘    │
+└─────────────────────────────────────────────────────────────┘
+                              │
+                              │ crossbeam channels
```
ReCore-sys: This isn't really true, crossbeam channels are used to communicate incoming packets back to the ecs, but the StreamWriter that is used to send outgoing packets from the ecs uses tokio channels

src/bin/src/systems/new_connections.rs
```diff
@@ -130,14 +131,6 @@ pub fn accept_new_connections(
         );
 
         trace!("Spawned entity for new connection: {:?}", entity_id);
-        // Add the new entity to the global state
```
ReCore-sys, line 133: Why was this removed?

src/bin/src/game_loop.rs
```diff
@@ -213,44 +327,48 @@ fn tcp_conn_acceptor(
 ) -> Result<(), BinaryError> {
     let named_thread = std::thread::Builder::new().name("TokioNetworkThread".to_string());
     named_thread.spawn(move || {
+        // Catch panics to ensure graceful shutdown even if something goes wrong
```
ReCore-sys, line 330: Should probably specify that we catch it so we can shut down the entire server instead of leaving it open with a crashed network loop

src/bin/src/launch.rs
```diff
@@ -0,0 +1,88 @@
...
+        start_time,
+    })
+}
+
+/// Generates spawn chunks around the origin if they don't exist.
+pub fn generate_spawn_chunks(state: GlobalState) -> Result<(), BinaryError> {
+    info!("No overworld spawn chunk found, generating spawn chunks...");
+
+    let start = Instant::now();
+    let radius = get_global_config().chunk_render_distance as i32 + 4;
```
ReCore-sys: We don't need the + 4, i think that was left over from a PR of mine, sorry

src/lib/net/src/conn_init/status.rs
```diff
@@ -192,7 +193,16 @@ fn get_server_status(state: &GlobalState) -> String {
...
 
     // Randomly choose a MOTD line from the configured list
-    let motd = config.motd.choose(&mut rand::rng()).unwrap();
+    const DEFAULT_MOTD: &str = "A FerrumC Server";
+    let motd: &str = config
+        .motd
+        .choose(&mut rand::rng())
+        .map(|s| s.as_str())
+        .unwrap_or_else(|| {
+            warn!("Add a MOTD line to your server config. Using default for now.");
```
ReCore-sys, line 202: We are using figment for the config so if the user doesn't specify one it'll fall back to whatever we have set in the default config

src/bin/src/systems/player_count_update.rs
ReCore-sys, line 1: Why did this get deleted?

Sweattypalms: Because we have the events already on player join/leave. Not sure why we need to recalculate every 10 seconds.

src/lib/net/src/conn_init/status.rs
Sweattypalms: What if its an empty array, wouldnt rand rng panic?

ReCore-sys: true, didn't think of that

src/bin/src/systems/player_count_update.rs
ReCore-sys: Mostly to prevent desyncs

Sweattypalms: so that means our event logic is dysfunctional? if that's the case, i fear we got bigger problems

src/bin/src/systems/new_connections.rs
Sweattypalms: Because it's literal duplicate from 125~131.

docs/protocol-flow.md
Sweattypalms: makes sense, fixed.

PR #284, opened 2025-11-30
introduce Position structs for the Chunk impl

Mcrtin:
This is a preview of how the chunk impl could look like.
TODO:
    - ~~current chunk gen tests aren't passing (so not that important xD)~~
    - i'd like to, instead of making changes on chunks, use chunk sections
    - ~~net positions should get integrated into the newly added position structs~~
    - the position structs need docs more functions and overall a glowup
    - Passing dims as strings sucks. (also, why do we not use &'static str instead of cloning?)
    - why do we use arcs in thu chunk impl?
    - why do we sanitize BlockStateId
    - ...
for now this has to suffice.a
the actual changes include:
- add position structs
- remove dim and pos from chunk

### DIFF
--- a/src/bin/src/launch.rs
+++ b/src/bin/src/launch.rs
@@ -45,18 +45,15 @@ pub fn generate_spawn_chunks(state: GlobalState) -> Result<(), BinaryError> {
     for (x, z) in chunks {
         let state_clone = state.clone();
         batch.execute(move || {
+            let pos = ChunkPos::new(x, z);
             let chunk = state_clone
                 .terrain_generator
-                .generate_chunk(ChunkPos::new(x, z))
+                .generate_chunk(pos)
                 .map(Arc::new);
 
             match chunk {
                 Ok(chunk) => {
-                    if let Err(e) =
-                        state_clone
-                            .world
-                            .save_chunk(ChunkPos::new(x, z), "overworld", chunk)
-                    {
+                    if let Err(e) = state_clone.world.save_chunk(pos, "overworld", chunk) {
                         error!("Error saving chunk ({}, {}): {:?}", x, z, e);
                     }
                 }
--- a/src/bin/src/packet_handlers/play_packets/player_action.rs
+++ b/src/bin/src/packet_handlers/play_packets/player_action.rs
@@ -53,7 +53,7 @@ pub fn handle(
                                 .0
                                 .clone()
                                 .terrain_generator
-                                .generate_chunk(BlockPos::from(event.location.clone()).chunk())
+                                .generate_chunk(pos.chunk())
                                 .map_err(BinaryError::WorldGen)?
                         }
                     };


NanCunChild:
;) remains a few bugs, cuz it is bug nest and hard to figure out what exactly happens.

src/lib/world/src/pos.rs
```diff
@@ -0,0 +1,301 @@
...
+use bevy_math::U8Vec2;
+use bevy_math::U8Vec3;
+use bevy_math::Vec2Swizzles;
+use bevy_math::Vec3Swizzles;
+use ferrumc_net_codec::net_types::network_position::NetworkPosition;
+
+#[derive(Clone, Copy)]
+pub struct BlockPos {
+    /// (i26, i12, i26)
+    pub pos: IVec3,
```
ReCore-sys, line 19: Would using the simd version be possible? Might allow for cleaner conversions from/to player positions

src/lib/world/src/pos.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+#[derive(Hash, Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ChunkPos {
+    pub pos: IVec2,
+}
+
+impl ChunkPos {
+    pub const fn new(x: i32, z: i32) -> Self {
+        assert!(x < 1 << 22);
```
ReCore-sys, line 116: What are these bitshifts doing?

src/lib/world/src/pos.rs
```diff
@@ -0,0 +1,301 @@
...
+use bevy_math::IVec3;
+use bevy_math::U8Vec2;
+use bevy_math::U8Vec3;
+use bevy_math::Vec2Swizzles;
+use bevy_math::Vec3Swizzles;
+use ferrumc_net_codec::net_types::network_position::NetworkPosition;
+
+#[derive(Clone, Copy)]
+pub struct BlockPos {
+    /// (i26, i12, i26)
```
ReCore-sys, line 18: Is using the packed format really worth it? I'd be worried about how that'd impact the CPU's native data type handling. This data type would largely be an intermediate bit of data so a few extra bytes won't make all that much of a difference.

src/lib/world/src/pos.rs
```diff
@@ -0,0 +1,301 @@
...
+    }
+}
+
+pub struct ChunkColumnPos {
+    pos: U8Vec2,
+}
+
+impl ChunkColumnPos {
+    pub const fn new(x: u8, z: u8) -> Self {
+        assert!(x < 16);
```
ReCore-sys, line 182: In non-debug builds should we just truncate down to 0-15?

src/lib/world/src/pos.rs
NanCunChild: Regarding the bitshifts (1 << 22), I believe this aligns with the previous discussion about using 22-bit widths for world coordinates to enforce boundaries and optimize storage.

src/lib/world/src/pos.rs
NanCunChild: I agree with @ReCore-sys on the packed format concern. For BlockPos which is used heavily in logic calculations, the CPU overhead of unpacking might outweigh the memory savings. Maybe we should keep the packed format only for storage/networking structs, but use native types for internal logic?

src/lib/world/src/pos.rs
Mcrtin: a Network position may only be at most 64 bit (divided between xyz) where x is at most 24 bit iirc so the chunk x pos may only be 22 bit. this is important for the pack() feature ChunkPos provides where we assume these bounds.

src/lib/world/src/pos.rs
Mcrtin: we aren't using the packed format... it just uses an ivec3 under the hood; the comment is just there to make the bounds clear, that we have to assure to be able to send it to the client via networkPosition

src/lib/world/src/pos.rs
Mcrtin: i was wondering if we should but i want const evaluations to error at comp time and i couldn't really figure out a better way

src/lib/world/src/pos.rs
Mcrtin: there is a simd version?

ReCore-sys: Ah my bad, I was thinking of the f32 version

PR #290, opened 2025-12-07
feat: allow player to swim

Tonguechaude:
Just allow player to swim ! 💦 (didn't find pool emoji sry)

image

image

### DIFF
--- a/src/bin/src/systems/player_swimming.rs
+++ b/src/bin/src/systems/player_swimming.rs
@@ -3,7 +3,7 @@ use bevy_math::DVec3;
 use ferrumc_components::player::swimming::SwimmingState;
 use ferrumc_core::identity::player_identity::PlayerIdentity;
 use ferrumc_core::transform::position::Position;
-use ferrumc_macros::block;
+use ferrumc_macros::match_block;
 use ferrumc_net::connection::StreamWriter;
 use ferrumc_net::packets::outgoing::entity_metadata::{EntityMetadata, EntityMetadataPacket};
 use ferrumc_net_codec::net_types::var_int::VarInt;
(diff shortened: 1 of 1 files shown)


src/bin/src/systems/player_swimming.rs
```diff
@@ -0,0 +1,84 @@
...
+    let eye_pos = IVec3::new(
+        pos.x.floor() as i32,
+        (pos.y + PLAYER_EYE_HEIGHT).floor() as i32,
+        pos.z.floor() as i32,
+    );
+
+    state
+        .world
+        .get_block_and_fetch(eye_pos.x, eye_pos.y, eye_pos.z, "overworld")
+        .map(|block_state| (86..=101).contains(&block_state.0))
```
GStudiosX2: you shouldn't hard code the water ids they can change in new updates

ReCore-sys: Agreed, we have a macro to resolve block IDs at compile time. It won't be as concise but you should probably use that

src/lib/components/src/player/swimming.rs
```diff
@@ -0,0 +1,21 @@
+use bevy_ecs::prelude::Component;
+
+/// Component tracking whether a player is currently swimming
+#[derive(Component, Debug, Clone, Copy, Default)]
+pub struct SwimmingState {
+    pub is_swimming: bool,
+}
+
+impl SwimmingState {
```
ReCore-sys: I don't think any of these methods are needed? The field is already public, just use regular assignment instead of getters/setters. You also already have a default impl so don't really need a `new()`

src/bin/src/systems/player_swimming.rs
```diff
@@ -0,0 +1,84 @@
...
+use ferrumc_net_codec::net_types::var_int::VarInt;
+use ferrumc_state::GlobalStateResource;
+use tracing::error;
+
+/// Height of player's eyes from feet (blocks)
+const PLAYER_EYE_HEIGHT: f64 = 1.62;
+
+/// Check if a player is in water by testing at eye level
+fn is_player_in_water(state: &ferrumc_state::GlobalState, pos: &Position) -> bool {
+    let eye_pos = IVec3::new(
```
ReCore-sys: Might be cleaner to put it all in a regular Vec3, then floor that and convert to IVec3

ReCore-sys (commented):
Mostly fine, just some OOP nitpicks

src/bin/src/systems/player_swimming.rs
Tonguechaude: Same as usual Vec3 use f32. So do you prefer I use a Vec3 and cast as f32 or use Dvec3 and use bevy_math method as_ivec3() ?

Position already use f64 so maybe Dvec3 is preferable ?

ReCore-sys: Yeah Dvec in that case, whatever matches the bit length

src/bin/src/systems/player_swimming.rs
```diff
@@ -0,0 +1,105 @@
...
+        block!("water", { level: 14 }),
+        block!("water", { level: 15 }),
+    ];
+
+    let pos = BlockPos::of(eye_pos.x, eye_pos.y, eye_pos.z);
+
+    state
+        .world
+        .get_block_and_fetch(pos, "overworld")
+        .map(|current_block| water_id.contains(&current_block))
```
ReCore-sys: `match_block!("water", current_block)`

PR #292, opened 2025-12-08
fix: `block` macro speedup

lexeyOK:
this pr adds optimisations and syntax improvements for the `block!` macro

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -244,6 +244,3 @@ criterion = { version = "0.7.0", features = ["html_reports"] }
 
 phf = { version = "0.13", features = ["macros"] }
 phf_codegen = { version = "0.13" }
-
-phf_codegen = { version = "0.13" }
-phf = { version = "0.13", features = ["macros"] }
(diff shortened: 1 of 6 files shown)


src/tests/src/block/mod.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(false)]
+mod test {
+    use ferrumc_macros::block;
+    #[derive(Debug, PartialEq, Eq)]
+    struct BlockStateId(u32);
+    #[cfg(false)]
```
ReCore-sys, line 6: ?

src/lib/derive_macros/src/lib.rs
```diff
@@ -119,6 +125,13 @@ pub fn build_registry_packets(input: TokenStream) -> TokenStream {
 ///
 /// If the block or properties are invalid, a compile-time error will be thrown that should hopefully
 /// explain the issue.
+/// # Static block
+/// gives a single block pat. Mainly used with
```
ReCore-sys, line 129: Could you add some examples with what you mean by this?

src/tests/src/block/mod.rs
lexeyOK: this was a test what uses all blockstates but you said i shouldn't add it since it's too long to compile, it's useful for me to check itacro is working but it also generates a lot of unnecessary functions so I disable it completely

src/lib/derive_macros/src/lib.rs
lexeyOK: i guess it's showing the distinction between having block with and without properties, if it has any properties they must be used or skiped explicitly, similar to how mattern matching works on structs I don't really know what to write here

PR #325, opened 2026-01-22
feature/block-behavior

MSKatKing:
This PR implements a simple system that allows for blocks to possess specific behavior such as different placement states, random ticking, and more.

I'm going to add methods that are still needed to the BlockBehavior trait, as well as adding documentation explaining how to implement behavior for a block. I am also going to rework the static registry to use block state ids rather than names.


src/bin/src/systems/handle_block_updates.rs
```diff
@@ -0,0 +1,81 @@
...
+                    z: update_pos.pos.z,
+                },
+                block_state_id: VarInt::new(updated_block as _),
+            };
+
+            packets.push((packet, update_pos));
+        }
+    }
+
+    let render_distance = get_global_config().chunk_render_distance;
```
ReCore-sys: Compare this with the player's render distance and choose the smallest

src/lib/blocks/crates/generated/src/blocks.rs
```diff
@@ -0,0 +1,336 @@
+mod bamboo_block;
```
ReCore-sys: Could we exclude these files from git? It's all generated a build time, don't see why we need to add a bunch of extra deltas for git to crunch every time we update versions

src/lib/blocks/benches/bench.rs
```diff
@@ -0,0 +1,17 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use ferrumc_blocks::BLOCK_MAPPINGS;
+
+fn benches(c: &mut Criterion) {
+    fn raw_func_dispatch_ctrl() {}
```
ReCore-sys: ??

src/lib/blocks/crates/build/src/complex.rs
```diff
@@ -0,0 +1,522 @@
...
+use ferrumc_block_properties::{PropertyDescriptor, TYPES};
+use heck::{ToPascalCase, ToShoutySnakeCase, ToSnakeCase};
+use proc_macro2::{Ident, TokenStream};
+use quote::{format_ident, quote};
+use std::collections::HashMap;
+
+struct BlockStateConfiguration<'a> {
+    name: &'a str,
+    properties: Vec<(&'a str, &'a str)>,
+    values: HashMap<u32, &'a HashMap<String, String>>,
```
ReCore-sys: idk how performance critical this is, but maybe something like fxhashmap if it's in a hot loop?

src/lib/blocks/benches/bench.rs
MSKatKing: I was comparing the time it took to call a function on a block to how long it took to call an empty function, I can remove this if you'd like.

ReCore-sys: I'm pretty sure the compiler would optimize that out, and I don't really see what that benchmark is for anyway. Probably remove it.

src/bin/src/systems/handle_block_updates.rs
```diff
@@ -0,0 +1,85 @@
...
+use ferrumc_net::packets::outgoing::block_update::BlockUpdate;
+use ferrumc_net_codec::net_types::network_position::NetworkPosition;
+use ferrumc_net_codec::net_types::var_int::VarInt;
+use ferrumc_state::GlobalStateResource;
+use ferrumc_world::block_state_id::BlockStateId;
+use ferrumc_world::pos::BlockPos;
+use tracing::error;
+
+#[derive(Resource, Default)]
+pub struct BlockUpdates(Vec<BlockPos>);
```
ReCore-sys, line 16: This should probably either be a VecDeque or a crossbeam queue

src/lib/blocks/src/behavior_trait.rs
```diff
@@ -0,0 +1,110 @@
...
+        _ret
+    };
+}
+
+/// Macro to autogenerate the `BlockBehavior` trait and associated VTable structs.
+///
+/// This macro simple exists to make adding methods to blocks easier.
+///
+/// The syntax for this macro is as follows: `fn <name>([mut]; <arguments>) [-> <return type>; <default return value>]`
+macro_rules! block_behavior_trait {
```
ReCore-sys, line 16: Could you give an example of how and where this would be used?

src/lib/blocks/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    include!(concat!(env!("OUT_DIR"), "/mappings.rs"));
+
+pub struct PlacementContext {
+    pub face: BlockFace,
+    pub cursor: DVec2,
+}
+
+impl BlockBehavior for SlabBlock {
+    #[inline(always)]
+    fn get_placement_state(&mut self, context: PlacementContext, world: &World, pos: BlockPos) {
```
ReCore-sys, line 29: How would this take into account stuff like placing blocks inside others (eg. trying to place a block on another block's face but there is a door/sign/trapdoor in the way)? It might also need to take into account stuff like entity collisions.

ReCore-sys (changes requested):

PR #342, opened 2026-03-23
Implement attack handler

ch-iv:
This PR enables attack events to be handled properly.

This pull request updates the decoding logic for the `InteractEntity` packet in the networking layer. It also adds a friction handler to make sure an entity slows down over time.

Note: I came up with velocity constants on my own. If you have a reference to an implementation in a more mature project, please let me know.

https://github.com/user-attachments/assets/038d1c42-d800-4d1b-b5f2-86c4bf207f6d


ch-iv:
I think the formatter is failing due to an inconsistency on main.

ahmed9959111-eng (approved):

PR #345, opened 2026-06-01
feat: add fluid simulation with parallel tick evaluation

NanCunChild:
feat: add fluid simulation with parallel tick evaluation

- ferrumc-core: add TickCounter resource (monotonic game tick counter)
- ferrumc-world: add fluid query table (water/lava level ↔ BlockStateId)
- ferrumc-world: add BlockTickScheduler (per-chunk scheduled tick queues,
  dedup, drain_due)
- ferrumc-world: add fluid spread algorithm (decoupled pure fn,
  BlockView trait, simplified model)
- bin: FluidScheduler ECS resource + process_fluid_ticks system
  (serial path for small batches, parallel path via ThreadPool for
  large batches, deterministic conflict reduction)
- bin: seed fluid ticks on block place and break (BlockBrokenEvent)
- bin: fix DashMap re-entrant write deadlock in place_block (drop
  chunk guard before seeding)
- BlockPos: derive Hash, Eq, PartialEq

(after many edition done, i have no idea what i have done. So i drive AI to make conclusions above)

### DIFF
--- a/src/bin/src/game_loop.rs
+++ b/src/bin/src/game_loop.rs
@@ -249,6 +249,7 @@ fn build_timed_scheduler() -> Scheduler {
     // Uses Burst behavior to catch up if ticks are missed (up to 5 at a time).
     let build_tick = |s: &mut Schedule| {
         s.set_executor_kind(ExecutorKind::SingleThreaded);
+        s.add_systems(crate::systems::tick_counter::handle); // Advance the global tick counter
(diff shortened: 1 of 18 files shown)


NanCunChild:
I noticed the problem of corrupting sections, it happens when fluid passing those blocks.
This problem occured 6mo ago, when we were digging, a section of blocks just gone.

NanCunChild:
correction, this problem is noticed.
something was off with palette conversions

PR #349, opened 2026-06-03
Fix section broken

NanCunChild:


NanCunChild:
[5ff9ad5](https://github.com/ferrumc-rs/ferrumc/pull/349/commits/5ff9ad54ca14b8b82a3751199806f4dce9083054)
This commit fixed new chunk no load problem, which due to:
- in-queue is differ from sending, so does both filters.
- in chunk calculation, calculater using min, but sender using max.

NanCunChild:
WARNING: I had only tested far away about 500 blocks, and all chunks loaded correctly. And I did not reproduce the "saving dropped" problem mentioned in discord.
I need more tests to prove this commit is safe

